Issue #41, opened 2015-01-01
D3-friendly outputs and templates


pbinkley:
My use case involves relatively small sets of tweets (~10k), from which I want to extract data to feed into various D3 visualizations: timelines, graphs, etc. I'll therefore be putting in a couple of PRs shortly (I hope), but some of this will stretch my python skills to the limit or beyond. I've got some work under way for parts of it but I don't lay claim to any of it.

- refactor the force-directed graph code to use a two-step process: a task-specific step to generate a json/csv data output, and a generic step to embed that output in a specified html template so that it's easy to get a quick look at your data. The data outputs would conform to the styles used most commonly in D3 examples, to make it easy to connect a given body of tweets to a given D3 example.
- (more speculative) refactor some of the current utilities to clarify the distinction between those that filter a tweet file (outputting a tweet file) and those that produce some other output, to make it easier to think in terms of pipelines. Call them filters and analyzers?
- add a filter to store a new field in each tweet of a tweet file. For one project I have a requirement to work with a local timezone rather than UTC, and it will be convenient to add the local time as a new field for further processing by other filters.
- add an analyzer to generate counts of co-occurrences of values in arbitrary fields. I've written one that will work specifically on hashtags (how many tweets in this set have #a and #b), which I feed into a D3 force-directed graph, but there's no reason not to make it generic to allow e.g. co-occurrence of mentioned users, or hashtags and mentioned users.
- some day: a core group of a few D3 visualizations that would be useful for any set of tweets to show (say) the temporal dimensions, histograms of users and hashtags, etc., that could be run easily to get an overview of a harvested body of tweets.

Comments and suggestions (and code!) welcome.

edsu:
This all sounds good to me @pbinkley so please send the pull requests. FWIW I've half considered creating a new project (twarc-report?) that is oriented around generating a cohesive static report (using d3, bootstrap, etc). If you want to run in that direction I would be willing to help, and it might give you space to think?

pbinkley:
I've put in PR #42 for the first item.
I wondered about a new project as well. For my purposes, I think the changes will manifest themselves as a series of utilities like directed.py, which analyze a given twitter file and build a dict or array of values, and then call a method in d3json.py to serialize the values in a format that can be dropped into a given D3 example. They'll also have an HTML template file for immediate viewing. To create a new project would mean moving some of what is currently in twarc/utils over, but that wouldn't be a big deal.
[...]

PR #42, opened 2015-01-04
Refactor directed.py to use d3json module

pbinkley:
Here's the first of the threatened PRs (in #41). The directed graph utility now uses a generic module to generate the D3-friendly json, and then optionally output it directly for reuse or embed it in the HTML template directed.html for immediate viewing. The new version is backwards-compatible, in that the default params produce the same output as the old version, but there are now options that can be set on the command line (via optparse). 

I hope that this provides a model for generating D3-friendly json in different formats, and I intend to build up d3json.py to do this. Next in line is a timeline bar chart.


edsu:
@pbinkley would it be possible to see this pull request with [spaces instead of tabs](http://legacy.python.org/dev/peps/pep-0008/#tabs-or-spaces)? It should be possible to configure your text editor to do this, but let me know if you run into problems.

edsu:
Thanks @pbinkley!

PR #49, opened 2015-01-25
Twarc report move

pbinkley:
Remove the d3 utils that have moved to the twarc-report project, and update README.md with information about twarc-report.


edsu:
Thanks for doing this in such a thorough manner @pbinkley! I should find some time to dig into twarc-report shortly!

PR #56, opened 2015-02-07
Add --use flag to read query from text file

pbinkley:
This allows you to keep your twarc query in a local text file, so that a given harvest might have a home directory in which the query and the harvested tweets live. The file makes it easier to manage more complex queries, and it makes the query available to twarc-report processes so that it can be automatically included in reports. 
It might also be useful to have a flag --save that would write a command-line query to twarc-search.txt. 
I've implemented it by overriding args.search with the string that is read from the file. I've tested with Cyrillic strings to be sure that Unicode is handled correctly.
I'm not sure how to handle archive.py so as to make it possible to use the --use flag without breaking its existing syntax. I'll have to read up on argparse options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@ each tweet to stdout as line oriented JSON. Twitter's search API only makes
 time is of the essence if you are trying to collect tweets for something 
 that has already happened. 
 
+You may also save the query in a text file, and load it using the --use flag:
+
+    twarc.py --use mysearch.txt > tweets.JSON
+
+Entering the --use flag without a file name will cause twarc to use "twarc-search.txt". 
+This flag is useful to keep a record of a complex query.
+
 ## Stream
 
 In stream mode twarc will listen to Twitter's [filter stream API](https://dev.twitter.com/streaming/reference/post/statuses/filter) for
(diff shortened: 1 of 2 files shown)


edsu:
Are you using archive.py then? 

Had you considered simply storing the twarc command in a file? So in your harvest directory you could have a file `command.sh` that looked like:

      twarc.py --query "some crazy complicated query here"

pbinkley:
I'm using both twarc.py and archive.py in different settings. The problem I'm trying to solve is that the query is used in two contexts: as part of a command to harvest and reharvest tweets, and as part of twarc-report's metadata about the collection of tweets from which it is generating a report. I'm using shell scripts for repeat harvests, and I could parse the query out of them for twarc-report, but it would be easier to make this a consistent feature of twarc-report if the query could be retrieved from the same place regardless of OS etc. But if you feel this doesn't belong in twarc then it's easy enough to make it a twarc-report-only feature (store the query in a text file, have command.sh read it from there so it can't get out of sync). I'm working towards an event-tracker script for twarc-report that will do periodic archive.py harvests for a given query and generate a suite of visualizations and post them online. The need for project-level metadata will probably go beyond just the query. So, maybe this is something that should simmer in twarc-report for a while before graduating (if ever) into twarc.

edsu:
Yeah, if it's ok I'd like to not have this be part of twarc.py Sorry if this is annoying but I just spent quite a bit of time simplifying it.

PR #57, opened 2015-02-19
Test utils on CI

hugovk:
Now twarc is Python 3 compatible, it may be useful to ensure the utils are as well.

This is a simple way to test them, just run them (via a shell script) on the CI. It doesn't verify the output (or include them in coverage), but does make sure they can be run on the CI Python versions.

For test data, it uses the first 100 IDs from ferguson-tweet-ids.zip and hydrates them.

Not all utils have been added yet, but this is a start. I've updated the ones that are run to be Py2/Py3 compatible.


hugovk:
Passing build on my branch https://travis-ci.org/hugovk/twarc/builds/51328767

edsu:
Nice, it would be super to run them as part of the test suite. Do you think that might be doable?  I will merge this for now to get started.

edsu:
Oh I see you do have the tests running via travis? Very nice.

hugovk:
Yep, these are also run on Travis. Not all the utils are in there, but they can be added gradually (and especially for new ones) to make sure they're compatible.

It wouldn't be too hard to get coverage for them too.

PR #61, opened 2015-03-03
added unicodecsv to requirements.txt

dchud:
this module is required for utils/json2csv.py but wasn't listed yet.


edsu:
Thanks @dchud!

PR #69, opened 2015-06-02
Fix downloads badge

hugovk:
pypip.in has been down for a while, but http://shields.io/ is another service. 

This is the monthly badge; change `dm` to `dw` or `dd` for weekly or daily downloads.

[CI skip]


edsu:
Thanks! :smiley_cat:

Issue #64, opened 2015-03-26
profiles


edsu:
It would be handy if twarc had a notion of profiles, similar to amazon's aws-cli. This way the various keys could be saved in a file, and there could be a default.

Wwhen started without any keys twarc could prompt for the consumer keys, give the user a URL to visit in their browser to get the access keys, and then save all the keys in their default profile ($HOME/.twarcrc).

hugovk:
I'm not familiar with aws-cli, but for my Twitter bots I use a (YAML) config file containing the four keys and secrets and pass that file as an argument.

For example:
https://github.com/hugovk/wotdbot/blob/master/data/wotdbot_example.yaml
https://github.com/hugovk/wotdbot/blob/master/wotdbot.py#L15-22

Would something like that be useful? Shall I make a PR?

edsu:
Yes, YAML would be great. A PR would be most welcome!

edsu:
This got done, but never got closed!

PR #71, opened 2015-06-03
Load Twitter keys/secrets from YAML config file

hugovk:
Re: https://github.com/edsu/twarc/issues/64

Add a `--yaml` command-line argument. If the Twitter keys and secrets haven't been found in an env variable or in their own arguments, then attempt to load them from this .yaml file.

Other things:

 * I removed the restriction to build only master, as its good practice to make PRs from branches, not master, and I wanted to test my changes on the CI. Let me know if you want the restriction putting back.

 * I put back the coverage and static analysis stuff in the build. This command had been causing problems:

```bash
coverage run --source twarc -m python setup.py test --verbose
No module named 'python'
```
The correct command is:
```bash
coverage run --source twarc setup.py test --verbose
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ json
 html
 custom
 img/
+*.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -64,4 +64,4 @@ json
 html
 custom
 img/
-*.yaml
+*.ini
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,17 @@ install:
 - python setup.py install
 
 script:
-- python setup.py test
+- pip install coverage
+- coverage run --source twarc setup.py test --verbose
 - chmod +x test/test_utils.sh
 - test/test_utils.sh # no coverage yet
 
+after_success:
+- pip install coveralls
+- coveralls
+
 after_script:
+- coverage report
 - pip install pep8 pyflakes
 - pep8 --statistics --count *.py
 - pyflakes *.py | tee >(wc -l)
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ python:
 - '2.7'
 - '3.4'
 
-branches:
-  only:
-  - master
-
 language: python
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,16 @@ install:
 
 script:
 - python setup.py test
+- chmod +x test/test_utils.sh
+- test/test_utils.sh # no coverage yet
+
+after_script:
+- pip install pep8 pyflakes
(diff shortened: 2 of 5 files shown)


edsu:
This looks like it is shaping up nicely. If no keys are found in the environment and --yaml hasn't been used do you think it would be useful if twarc prompted the user for the keys and saved them to some default location for the yaml file, like ~/.twarc?

hugovk:
Good idea, I'll update the PR.

hugovk:
PR updated. Keys are prompted for and saved to `~/twarc.yaml`. Would you prefer `~/twarc`?

Perhaps the default location for the `--yaml` option should be updated to match.

edsu:
~/.twarc.yaml works too -- thanks!

hugovk:
Updated `--yaml` default to match. Also flake8 changes.

hugovk:
Build passed on branch: https://travis-ci.org/hugovk/twarc

edsu:
This is great. When updating my own dev environment I see that pyyaml requires compiling C libraries. This seems like a bit of a pain honestly. Is there any reason not use something that comes with Python: json or configparser, etc? Or am I being overly conservative here?

hugovk:
At least for me `pip install pyaml` was painless on OS X and Windows. I didn't notice if it built anything. But if the installation doesn't go smoothly on other platforms, I can certainly change it to JSON or ConfigParser. 

One good thing about YAML over JSON is the output is much cleaner, it's easier to create a valid file by hand; and it's a little bit simple than ConfigParser.

Here's a comparison: http://martin-thoma.com/configuration-files-in-python/

What do you think?

edsu:
If it's ok with you I'd like to switch it over to [ConfigParser](https://docs.python.org/2/library/configparser.html). I like the fact that it's core Python and easily supports multiple sections which could correspond to profiles as it works with [aws-cli](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-multiple-profiles). This will make it easy to run twarc using different credentials as needed ; which can be handy for concurrent archiving jobs.

That being said, I don't expect you to do this if you don't want to. I can run with what you have if you've pushed it as far along as you can right now.

hugovk:
Sounds good, I'm happy to switch it. I'll do it in the next few days.

hugovk:
Updated with ConfigParser, build passes on my repo:
https://travis-ci.org/hugovk/twarc/builds/65672474

edsu:
Sorry for the delay, this is really, really super @hugovk -- thanks for the contribution!

:cake: :balloon: :tada:

hugovk:
Happy to help!

PR #78, opened 2015-09-30
add optional lang param to search

miku:
Twitter supports a few languages and has support for this in the API.

* https://dev.twitter.com/rest/reference/get/help/languages
* https://dev.twitter.com/rest/reference/get/search/tweets

> lang (optional): Restricts tweets to the given language,
  given by an ISO 639-1 code. Language detection is best-effort.

----

Alternative: Since [search](https://dev.twitter.com/rest/reference/get/search/tweets) supports even more optional parameters, it would maybe more general to write something like this:

    def search(self, q, max_id=None, since_id=None, **kwargs):
        ...
        params.update(kwargs)
        ...

Do you have any preferences?

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -233,10 +233,10 @@ class Twarc(object):
         self.access_token_secret = access_token_secret
         self._connect()
 
-    def search(self, q, max_id=None, since_id=None):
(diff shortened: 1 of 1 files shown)


edsu:
Thanks for this! Let's start with lang and add others as needed.

PR #91, opened 2016-03-28
added `result_type` option to `search()` with default value of `recen…

dchud:
…t`; removed unused import

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -2,7 +2,6 @@
 
 from __future__ import print_function
 
-import io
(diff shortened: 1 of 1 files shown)


twarc.py
```diff
@@ -2,7 +2,6 @@
 
 from __future__ import print_function
 
-import io
```
edsu: Thanks!

edsu:
Thanks @dchud!

PR #95, opened 2016-04-10
Create valid GeoJSON

alexgleith:
Reverted the geometry storage to ensure valid GeoJSON is being stored.

### DIFF
--- a/utils/geojson.py
+++ b/utils/geojson.py
@@ -60,6 +60,7 @@ features = []
 
 for line in fileinput.input(files=args.files):
     tweet = json.loads(line)
+    hasGeo = False
 
     f = {
         "type": "Feature",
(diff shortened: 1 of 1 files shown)


edsu:
Thanks, but it's not needed.

edsu:
I just caught up with some of your other messages and see why you want to have this. Sorry!

alexgleith:
Ok, well you're the boss, but I think it's a mistake not using standard GeoJSON. (Ah, now I see your edit!)

And, I seem to have missed the second commit, but I think there's a bug in that all features are getting passed through this utility now, rather than only the ones with a location.

https://github.com/alexgleith/twarc/blob/bounding-boxes-valid-geojson/utils/geojson.py

edsu:
Sorry for the confusion. I now see your point about the Polygon and the geo-less tweets being passed through. I think this has been resolved now?

alexgleith:
Yep, I think we're all good!

edsu:
Thanks for taking the time to contribute this, and for your patience :-D

alexgleith:
No worries, I'm keen to do more contribution to things like this rather than changing my local version and ignoring the main repo! Thanks for building the project =)

PR #96, opened 2016-04-11
Only include geo tweets

alexgleith:


alexgleith:
Sorry, this is a bit messy... I'm an amateur at the Git stuff! The second commit is important.

alexgleith:
Actually, I think this is incorporated in #95 .

PR #97, opened 2016-04-11
Fix JS link

hugovk:
Although the wordcloud is showing fine even though the old link is 404, that's because the JS is duplicated in this file.

Either the link or the JS code should be removed.

### DIFF
--- a/utils/wordcloud.py
+++ b/utils/wordcloud.py
@@ -54,7 +54,6 @@ print(("""<!DOCTYPE html>
 <meta charset="utf-8">
 <title>twarc wordcloud</title>
 <script src="http://d3js.org/d3.v3.min.js"></script>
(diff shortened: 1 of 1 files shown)


edsu:
Do you want to remove the embedded JavaScript then?

hugovk:
Just tried that (locally) and it didn't was just blank. Console said:

> Refused to execute script from 'https://raw.githubusercontent.com/jasondavies/d3-cloud/master/build/d3.layout.cloud.js' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
> Uncaught TypeError: d3.layout.cloud is not a function

So I removed the link instead.

edsu:
Ahah, that's why I added that workaround before but didn't clean up afterwards. Thanks!

PR #100, opened 2016-04-11
Add -v/--version flag to return the version number

### LINKED ISSUE #99: Add a -v flag for twarc to return the version number
What the title says.

I can take care of this in the next day or so if you want @edsu.

hugovk:
Fixes #99.

```bash
$ twarc.py --version
twarc 0.6.1
$ twarc.py -v
twarc 0.6.1
```

Put the version number in version.py and import that into setup.py and twarc.py.

It can't be put in twarc.py and imported into setup.py, because at that time the requests module (imported by twarc.py) may not have been installed yet by setup.py.

See http://stackoverflow.com/a/24517154/724176

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ html
 custom
 img/
 *.ini
+.eggs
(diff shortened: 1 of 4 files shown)


edsu:
Does `version.py` get installed when you `pip install twarc`? I'm a bit confused about how importing version works once it's installed.

hugovk:
You're right, version.py isn't installed. Do you know the best way to do so?

At least the very simplest thing would be to just duplicate the version number variable, but both would need updating for new releases.

PR #104, opened 2016-04-18
Fix typo in README.md

ianmilligan1:
@ruebot and I were doing final proofs in our _Code4Lib_ draft (which uses twarc) – we caught this typo in one of our block quotes. Thought this would be more useful than using [sic].


edsu:
Cheers!

PR #107, opened 2016-06-12
Add --geocode option for location-based searches

chosak:
This PR adds a `--geocode` command line option that maps to the [geocode](https://dev.twitter.com/rest/reference/get/search/tweets#api-param-search_geocode) search API option. It supports location-based searches with or without a query parameter.

It also adds a new `utils/geofilter.py` utility to allow for filtering of tweets by presence/absence of geo data, as well as filtering by arbitrary GeoJSON shape (e.g. a polygon).

`README.md` has been updated and all code has been tested with both Python 2.7 and Python 3.5.

**New `--geocode` command line option**

The search API `geocode`parameter lets you retrieve tweets from within a given radius (in miles or kilometers) from a given location, specified as latitude and longitude (in degrees). For example:

```sh
twarc.py --search ferguson --geocode 38.7442,-90.3054,1mi
```

You can also omit the search term if you want to retrieve all available tweets from the area:

```sh
twarc.py --geocode 38.7442,-90.3054,1km
````

I added a new test as `test_twarc.test_geocode` reusing the logic from `test_twarc.test_locations`.

**New `utils/geofilter.py` utility**

Using the `geocode` search parameter can pull back tweets with different kinds of geo data: the specific lat/lng coordinates it was sent from, a [Place](https://dev.twitter.com/overview/api/places) it was associated with (although potentially not sent from), or neither of these, returned only because the user's profile location is a match.

A new `utils/geofilter.py` utility lets you filter a file of JSON tweets by presence/absence of both coordinate and Place data. Note that you can both pass the filename in as a command line parameter or pipe a stream via stdin:

```sh
utils/geofilter.py tweets.json --yes-coordinates > tweets-with-geocoords.json
cat tweets.json | utils/geofilter.py --no-place > tweets-with-no-place.json
```

This utility also lets you optionally filter tweets by a more complex geofence (using an additional [Shapely](https://github.com/Toblerity/Shapely) requirement). For example, consider the use case where you want to find tweets from a certain neighborhood for which you have a GeoJSON outline (as a Polygon). Using this new utility, you could do an initial search for your keyword of interest to collect all tweets and then filter them using:

```sh
utils/geofilter.py all-tweets.json --fence limits.geojson > fenced-tweets.json
```

This is also useful when used in combination with the new `--geocode` option, e.g. first finding tweets near some point and then filtering that circular search area down to something more specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,15 @@ each tweet to stdout as line oriented JSON. Twitter's search API only makes
 time is of the essence if you are trying to collect tweets for something
 that has already happened.
 
+### Search for tweets within a given area
+
+You can specify a search term or omit one to search for all tweets within a given radius of a given latitude/longitude:
+
+    twarc.py --search ferguson --geocode 38.7442,-90.3054,1mi
+    twarc.py --geocode 38.7442,-90.3054,1mi
+    
+See the [API documentation](https://dev.twitter.com/rest/reference/get/search/tweets#api-param-search_geocode) for more details on how these searches work.
+
 ## Filter Stream
 
 In filter stream mode twarc will listen to Twitter's [filter stream API](https://dev.twitter.com/streaming/reference/post/statuses/filter) for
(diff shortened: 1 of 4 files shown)


edsu:
Ouch, I'm sorry I didn't see this till now! I will review ASAP.

PR #108, opened 2016-07-08
Added commandline configuration of pool size and unshrtn url to unsho…

justinlittman:
…rten.py.


edsu:
Thanks @justinlittman!

PR #109, opened 2016-07-11
Added configurable retrying of unshrtn service.

justinlittman:
To handle cases where unshrtn crashes.

Combination of retrying and adding --restart:always seems to handle crashes.


edsu:
Thanks! It's been a while since I looked at unshrtn, so if you notice it crashing definitely add an issue for that.

PR #110, opened 2016-07-14
Optionally read tweet id’s from the standard input during hydration.

dimazest:
This is a proof of concept implementation.

The good points:

* it can read archives
* it also reads from the std input.

The not so good points:

* the code is a bit ugly because I tried to eliminate an inline `open()`.

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 
+import fileinput
 import os
 import sys
 import json
@@ -140,7 +141,7 @@ def main():
         tweets = t.filter(track=args.track, follow=args.follow,
                           locations=args.locations)
     elif args.hydrate:
-        tweets = t.hydrate(open(args.hydrate, 'rU'))
+        tweets = t.hydrate(args.hydrate)
     elif args.sample:
         tweets = t.sample()
     elif args.timeline:
@@ -319,7 +320,7 @@ def catch_timeout(f):
 
 def catch_gzip_errors(f):
     """
-    A decorator to handle gzip encoding errors which have been known to 
+    A decorator to handle gzip encoding errors which have been known to
     happen during hydration.
     """
     def new_f(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


edsu:
This is beautiful, thank you! The only problem I'm having is how do you get twarc.py to read the ids from stdin? 

```
cat ids.txt | twarc.py --hydrate > tweets.json
usage: twarc [-h] [-v] [--search SEARCH] [--max_id MAX_ID]
             [--since_id SINCE_ID] [--result_type {mixed,recent,popular}]
             [--lang LANG] [--track TRACK] [--follow FOLLOW]
             [--locations LOCATIONS] [--sample] [--timeline TIMELINE]
             [--timeline_user_id TIMELINE_USER_ID]
             [--lookup_screen_names LOOKUP_SCREEN_NAMES [LOOKUP_SCREEN_NAMES ...]]
             [--lookup_user_ids LOOKUP_USER_IDS [LOOKUP_USER_IDS ...]]
             [--hydrate HYDRATE] [--log LOG] [--consumer_key CONSUMER_KEY]
             [--consumer_secret CONSUMER_SECRET] [--access_token ACCESS_TOKEN]
             [--access_token_secret ACCESS_TOKEN_SECRET] [-c CONFIG]
             [-p PROFILE] [-w]
twarc: error: argument --hydrate: expected one argument
```

edsu:
Oh I see now: 

    cat ids.txt | twarc.py --hydrate - > tweets.json

twarc.py
```diff
@@ -140,7 +141,16 @@ def main():
         tweets = t.filter(track=args.track, follow=args.follow,
                           locations=args.locations)
     elif args.hydrate:
-        tweets = t.hydrate(open(args.hydrate, 'rU'))
+        f_names = [args.hydrate] if args.hydrate != '--hydrate' else None
```
edsu: @dimazest can you tell me what this line is doing?

edsu:
i made a few changes and pulled them into master. Thanks for pushing this forward!

twarc.py
dimazest: The idea was to read from stdin if `-` is passed to pipe ids into twarc:

```
zcat ids.csv.gz | twarc.py --hydrate -
```

but in this case `args.hydrate` will somehow be equal to `--hydrate`.

edsu: I wee. Well I've tested the use of '--hydrate -' with what is now on master and it works fine. Thanks very much for your contribution @dimazest

PR #114, opened 2016-09-17
Added support for using unicode characters

Segerberg:
Added support for using unicode characters as command line arguments

### DIFF
--- a/utils/twarc-archive.py
+++ b/utils/twarc-archive.py
@@ -44,9 +44,9 @@ archive_file_pat = "tweets-(\d{4}).json$"
 def main():
     e = os.environ.get
     parser = argparse.ArgumentParser("archive")
-    parser.add_argument("search", action="store",
+    parser.add_argument("search", type=commandline_unicode, action="store",
                         help="search for tweets matching a query")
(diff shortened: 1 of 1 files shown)


utils/twarc-archive.py
```diff
@@ -44,9 +44,9 @@
 def main():
     e = os.environ.get
     parser = argparse.ArgumentParser("archive")
-    parser.add_argument("search", action="store",
+    parser.add_argument("search", type=commandline_unicode, action="store",
```
edsu: I see this when running after applying these changes:

NameError: global name 'commandline_unicode' is not defined

Segerberg: I'm sorry, my bad. Made a stupid mistake not including a needed function. 


#Function to decode args to unicode
def commandline_unicode(bytestring):
    unicode_string = bytestring.decode(sys.getfilesystemencoding())
    return unicode_string

edsu:
Sorry to have dropped the ball on this. I just tested the latest released twarc with python 3 and it works just fine on the command line, for example:

```
% twarc search 'ديربي_جده' > tweets.json
```

Are you on 2.7 still?

PR #123, opened 2016-12-23
Skip failing tests on PRs

hugovk:
**Not yet ready for merge; still testing.**

Right now, PRs fail on the CI, most likely because they don't have Twitter credentials as secure environment variables.

It would be useful if the CI could still at least do _some_ tests for PRs, rather than fail all the time. That way, PR CI failures would be something to pay attention to (e.g. something has broken), rather than something to ignore.

This refactors to run the tests using `unittest` so `unittest.skipIf` can be used.

This PR isn't yet ready for merge, I need to create a PR to see this would actually run as a PR! :)

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -36,6 +36,21 @@ T = twarc.Twarc(consumer_key, consumer_secret,
                 access_token, access_token_secret)
 
 
+def _travis_ci_but_no_secure_env_vars():
+    """Helper function to say if we're testing on Travis CI,
+    but don't have secure env vars
+    """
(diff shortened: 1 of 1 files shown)


edsu:
I would rather get tests to work on PRs instead of skipping them.

edsu:
Sorry this PR is now quite stale. Hopefully we can reopen it if a way to get tests working on branches is found?

One alternative would be for me to give you rights to the repo (since you've helped out so much) and also to have tests only run on the master branch.

PR #129, opened 2017-01-13
Python 3.6 is out

hugovk:
We're also probably more interested in Python 3.6 pass/fail before than Python 3.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 python:
 - '2.7'
-- '3.4'
+- '3.6'
 - '3.5'
+- '3.4'
 language: python
 install:
 - python setup.py install


edsu:
Do you think checking all the 3.x versions is necessary? How about just the latest?

hugovk:
It'd certainly be quicker to just test the latest.

It partly depends what you want to support, but Python 3.4 to 3.6 are probably close enough not to make much of a difference if any.

Python 3.2 has some big differences, but you're not testing that and it's not supported anyway: https://en.wikipedia.org/wiki/CPython#Version_history.

edsu:
Ok, I guess it's just Travis that takes longer. Tests will be just the same for development. Thanks again!

PR #131, opened 2017-01-14
Avoid HTTPS certificate error

hugovk:
Visiting https://www.docnow.io results in a certificate error:

image

image

image

Use http://www.docnow.io instead.


edsu:
www.docnow.io really needs to be available over HTTPS. But it's a Github pages site so I need to figure that out, or just run a server with a certificate. This helps in the meantime, thanks!

PR #133, opened 2017-01-14
added lockfile with pid of running process

### LINKED ISSUE #85: twarc-archive.py lockfile
It would probably be useful if twarc-archive.py created a lockfile in the target directory to prevent multiple processes from writing to the location. This would prevent a cron job from kicking off a long running harvest and then starting another one later.

nwoodward:
Fixes #85. If it would be useful, I could add a check to see if the pid in the lockfile is stale, and if so then rewrite the lockfile with a new pid and continue on with the current process.

### DIFF
--- a/utils/twarc-archive.py
+++ b/utils/twarc-archive.py
@@ -76,7 +76,7 @@ def main():
         format="%(asctime)s %(levelname)s %(message)s"
     )
 
-    lockfile = os.path.join(args.archive_dir, '') + ".lockfile"
+    lockfile = os.path.join(args.archive_dir, '') + "lockfile"
     if not os.path.exists(lockfile):
         pid = os.getpid()
         lockfile_handle = open(lockfile, "w")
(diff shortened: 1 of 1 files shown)


edsu:
Thanks @nwoodward! I wonder, do you think the lockfile should be visible in the archivedir? As a dotfile it might go unnoticed.

If the error mentions that the lockfile might be stale, and its location, that could get around having to add logic for handling it?

nwoodward:
Sure, I'll remove the dot. 

In order to check that it's stale I'd search for it in a list of current pids. The best way I've found to get such a list is with psutil, but that would be adding a dependency. Are you cool with that, or do you have another approach in mind?

https://pypi.python.org/pypi/psutil/5.0.1

nwoodward:
Ah, never mind. I see now you just wanted a message about it. I'll change that.

edsu:
Thanks @nwoodward!

Issue #119, opened 2016-11-13
Keep Alive tracking


edsu:
I've had a few situations where  week or month long filter tracking processes eventually go quiet. It would be useful if twarc.py monitored the stream and if it hasn't received a Keep Alive messages. If no keep alive is seen then the connection should be re-opened.

igorbrigadir:
Fixed in #234

PR #138, opened 2017-01-17
Add event parameter to streams to allow interrupting

justinlittman:
When used as a library, it is helpful to be able to gracefully interrupt streams.  This adds a parameter in which a threading.event() can be provided.  When provided and triggered, streams will be interrupted.

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -592,6 +592,8 @@ class Twarc(object):
                 for line in resp.iter_lines(chunk_size=1024):
                     if event and event.is_set():
                         logging.info("Stopping filter")
+                        # Explicitly close response
+                        resp.close()
                         return
                     if not line:
                         logging.info("keep-alive")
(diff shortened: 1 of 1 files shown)


edsu:
I'd be curious to hear what prompted you to want this. Can't the code that is reading from the stream decide to stop whenever it wants?

justinlittman:
On high velocity filters, yes.  However, on low velocity streams the client
code that is reading the iterator blocks until a tweet is available.
Within the Twarc code, the event can interrupt on a keep-alive, which
according to the Twitter docs should be at least every 30 seconds.

edsu:
perhaps just an option to pass along the keep-alives would have worked?

edsu:
Although I think I've noticed situations where the stream goes completely dead on occasion, probably due to network failure of some kind #119. So this approach of passing along an event could work nicely for situations like that.

justinlittman:
That would have worked.  However, I'd suggest that isolating the client
from the API by just returned tweets is (slightly) preferable.

PR #139, opened 2017-01-22
Includes the parent (root) tweet when recursive is True

paulgb:
I think it makes sense to include the root tweet when called with `--recursive`. I'm also happy to make this its own flag if that's preferable.


edsu:
I've wanted that too, thanks!

PR #151, opened 2017-03-04
Remove duplicate "user_screen_name"

irfreitas:
The get_row() and get_heading() functions each have "user_screen_name" attribute included twice, which can cause problems when importing the resulting csv to pandas later.


edsu:
Thanks for catching that!

PR #154, opened 2017-03-17
Adding timeline command to twarc-archive.py

toddstoffer:
I have added a twarc_command argument to twarc-archive.py. This allows users to pass either the search or timeline command to the twarc-archive.py script. If no additional argument is passed, it defaults to search.

### DIFF
--- a/utils/twarc-archive.py
+++ b/utils/twarc-archive.py
@@ -5,12 +5,12 @@ This little utility uses twarc to write Twitter search results to a directory
 of your choosing. It will use the previous results to determine when to stop
 searching.
 
-So for example if you want to search for tweets mentioning "ferguson" you can 
+So for example if you want to search for tweets mentioning "ferguson" you can
 run it:
 
     % twarc-archive.py ferguson /mnt/tweets/ferguson
 
-The first time you run this it will search twitter for tweets matching 
+The first time you run this it will search twitter for tweets matching
 "ferguson" and write them to a file:
 
     /mnt/tweets/ferguson/tweets-0001.json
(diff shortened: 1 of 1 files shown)


edsu:
Thanks for sending this along. When I try to run it like this:

    ./twarc-archive.py --twarc_command timeline edsu edsu

I get an edsu directory that contains an archive.log but there are no tweets in there. I will investigate but I'm curious if you are seeing the same thing?

toddstoffer:
I have been running it like this. ./twarc-archive.py toddstoffer
toddstoffer --twarc_command timeline

I think it has to do with how the twarc-archive.py is structured with the
first two arguments being positional (search, archive_dir) and the rest
being optional.

edsu:
@toddstoffer unfortunately that doesn't seem to work for me:

```
%  ./twarc-archive.py toddstoffer toddstoffer --twarc_command timeline
% ls -l toddstoffer/
total 24
-rw-r--r--  1 ed  staff  8372 Mar 17 16:07 archive.log
```

toddstoffer:
Sorry. I forgot that in production we have been using the id for timeline
harvests instead of a username (as usernames can change over time but ids
will not). So I was testing it with the following.

./twarc-archive.py 29541732 toddstoffer --twarc_command timeline

I will dig a little deeper this weekend to try and figure out why it isn't
harvesting when passing a username.

edsu:
Thanks, that was enough to narrow it down. I just needed to change the way Twarc.timeline was invoked depending on whether a screen_name or user_id is passed in.

Thanks again for taking the time to share back what you've been using!

PR #156, opened 2017-03-31
Set tweet mode when posting during hydrating.

justinlittman:
Steps to reproduce:
```
 echo "847526347039789057" | python twarc.py --tweet_mode extended hydrate - | jq
```

Expected result:
An extended tweet

Actual result:
A classic tweet

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -727,7 +727,7 @@ class Twarc(object):
             ids.append(tweet_id)
             if len(ids) == 100:
                 logging.info("hydrating %s ids", len(ids))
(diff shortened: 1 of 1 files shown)


edsu:
Actually, this is what I see:

```
echo "847526347039789057" | twarc --tweet_mode extended hydrate - | jq .full_text
"@mistersql From https://t.co/wAwuKkfaMv https://t.co/OUv3XnS4kM"
```

Are you sure you have twarc v1.0.10?

justinlittman:
I'm sure I'm not using v1.0.10 because then everything would have worked properly. Sorry for the confusion.

Issue #149, opened 2017-02-21
log authorization errors


cazzerson:
If a timeline search for a protected or deleted account is run, an exception is generated that prevents the lockfile from being deleted. Any future run will fail due to the existence of the lockfile. For example:

    requests.exceptions.HTTPError: 401 Client Error: Authorization Required for url: https://api.twitter.com/1.1/statuses/user_timeline.json?count=200&user_id=479065673

edsu:
Thanks, this seems fixable. What would you want to see happen? An error in the log and continuing on like normal?

cazzerson:
@edsu That sounds about right to me. Thanks!

edsu:
Just for clarification, are you getting this error when you run `utils/twarc-archive.py`?

cazzerson:
@edu Yes, that's right.

edsu:
@cazzerson how do you fetch a user timeline with twarc-archive.py? I thought it only did search?

cazzerson:
@edsu Ack, sorry about that. This is a locally modified version of the archive script :-) Let me transfer this to @bretdavidson.

edsu:
It sounds like some functionality that others might find useful...

bretdavidson:
@edsu @cazzerson Yeah, sorry about the confusion. We just put this together in the past few weeks. Will try to open a PR soon. Thanks for looking at this!

igorbrigadir:
I think this can be closed as the functionality is provided by https://github.com/DocNow/twarc-timeline-archive now?

PR #157, opened 2017-04-05
Adding additional http codes to timeline exception

toddstoffer:
This is a proposed fix for issue #149 . I have added additional http response codes to the timeline harvest exception. This extends the current behavior of 404 codes to additional status codes that represent protected accounts or 5xx codes in case of a Twitter outage. The end result is that twarc-archive.py can run without raising an exception, allowing lockfiles to be removed from all directories once the script completes.

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -466,8 +466,8 @@ class Twarc(object):
             try:
                 resp = self.get(url, params=params, allow_404=True)
             except requests.exceptions.HTTPError as e:
-                if e.response.status_code >= 400:
-                    logging.info("HTTP Response: %s - No timeline available for user %s",e.response.status_code,id)
+                if e.response.status_code in [401, 404, 502, 503, 504]:
+                    logging.info("HTTP Response: %s - No timeline available for user %s", e.response.status_code, id)
                     break
                 raise e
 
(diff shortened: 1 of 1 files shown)


twarc.py
```diff
@@ -466,8 +466,8 @@ def timeline(self, user_id=None, screen_name=None, max_id=None,
             try:
                 resp = self.get(url, params=params, allow_404=True)
             except requests.exceptions.HTTPError as e:
-                if e.response.status_code == 404:
-                    logging.info("no timeline available for %s", id)
+                if e.response.status_code in [401, 404, 502, 503, 504]:
```
edsu: Hi Todd, have you actually seen 502, 503 and 504 errors coming back from the API?

toddstoffer:
Hi @edsu I have not seen it yet, but could see a situation in which our cron job is running and Twitter goes down for a bit, which would send back one of the 5xx errors and cause a lockfile to be created. What are your thoughts?

edsu:
I noticed that a unit test is failing with this commit so I'll need to think it through a little bit. It feels to me like the exception handling belongs in twarc-archive.py (since that manages the lock file) and not in twarc.py itself. But maybe I'm mistaken.

edsu:
Unfortunately this PR is pretty stale now. I'm sorry it wasn't able to merged in a timely fashion.

PR #162, opened 2017-04-26
For streams, makes sleeping interruptible.

justinlittman:
Both the filter and sample methods support providing an event used to interrupt the stream. However, the event doesn't interrupt sleep. This changes sleep to a loop to allow checking the event and interrupting if appropriate.

### DIFF
--- a/twarc.py
+++ b/twarc.py
@@ -369,6 +369,21 @@ def catch_gzip_errors(f):
     return new_f
 
 
+def interruptible_sleep(t, event=None):
+    """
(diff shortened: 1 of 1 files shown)


edsu:
This is starting to feel a bit hokey, but I'll merge it in since it seems to be important for however you are managing streams. If things need to get any more complicated I think we'll need to revisit what's going on here.

edsu:
This was just released to PyPI as v1.1.1.

justinlittman:
I'm with you on this. One of the few things I miss about Java is the
ability to interrupt threads.

PR #175, opened 2017-08-22
Update docs to use .jsonl instead of .json

### LINKED ISSUE #173: documentation should use .jsonl extension
hey folks - love the tool, wondering why it doesn't create valid json?  i've looked at it in various validator tools and it always throws errors since there are multiple root elements.   how do other people resolve this or is it a non-issue for some reason

thanks again - great tool

hugovk:
Fixes #173 to make it clear twarc deals in line-oriented JSON rather than pure JSON.

Changed READMEs and utils.

Should either of these be changed or added to?

* https://github.com/DocNow/twarc/blob/master/utils/network.py#L156
* https://github.com/DocNow/twarc/blob/master/utils/extractor.py#L41

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ twarc
 twarc is a command line tool and Python library for archiving Twitter JSON data.
 Each tweet is represented as a JSON object that is
 [exactly](https://dev.twitter.com/overview/api/tweets) what was returned from
-the Twitter API.  Tweets are stored as line-oriented JSON.  Twarc will handle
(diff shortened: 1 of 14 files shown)


edsu:
Thanks @hugovk! I think network.py is fine since that is outputting a valid JSON document. I actually am embarrassed to admit that I've never seen extractor.py before! It does look like it should probably be updated too.

hugovk:
Wherever it came from, extractor.py is now updated!

edsu:
Many thanks for this @hugovk and @alibama for raising the issue.

PR #183, opened 2017-10-02
Adds tweet_compliance to utils.

justinlittman:
Supports [tweet compliance](https://developer.twitter.com/en/docs/tweets/compliance/overview).
That is, providing the most recent version of a tweet or removing unavailable (deleted or protected)
tweets.

Also useful for splitting out available tweets from unavailable tweets.

### DIFF
--- a/utils/tweet_compliance.py
+++ b/utils/tweet_compliance.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+
+"""
+Supports tweet compliance. See https://developer.twitter.com/en/docs/tweets/compliance/overview.
+That is, providing the most recent version of a tweet or removing unavailable (deleted or protected)
+tweets.
+
+Also useful for splitting out available tweets from unavailable tweets.
+
+Example usage: python tweet_compliance.py test.txt > test.json 2> test_delete.txt
+
+For each tweet in a list of tweets or tweet ids provided by standard input or contained in files,
+looks up the current tweet state.
+
+If a tweet is not available and tweet ids are provided, the tweet id is
+output to standard error.
+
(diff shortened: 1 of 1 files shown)


edsu:
This is so similar to hydrate, I wonder if it might make sense to add the logic of keeping track of missing tweets, and logging them, to twarc proper?

justinlittman:
I thought about that -- they're close. However, the specific problem I was trying to solve was splitting out deleted tweets from a collection of tweets. Thus, the input was a collection of tweets and the output was a collection of updated tweets and a collection of deleted tweets.

I couldn't come up with a way of doing that without making hydrate much more complicated. But definitely open for suggestions.

edsu:
But it looks like the input is a file of tweet ids, not a file of tweets. I think it would be nice for twarc itself to indicate in the log which ids had been deleted. But I guess you have some workflow that is expecting a file of deleted tweet ids and not a twarc log?

justinlittman:
Actually it takes a collection of tweets OR tweet ids. If tweets, then deleted tweets are outputted to STDERR. If tweet ids, then deleted tweet ids are outputted to STDERR. In both cases, the updated tweets are output to STDOUT.

In my case, I'm interested in the deleted tweets (not just their tweet ids).

edsu:
Ok, that is kinda complicated and doesn't fit so easily into twarc itself. I think I will add similar logic for reporting deleted tweet ids, since I've wanted that in the past.

justinlittman:
Yeah, my intuition was that it should go in hydrate but came to the
conclusion that it would overload hydrate.

FYI -- Motivation is for a researcher that wants to look at deleted
election tweets. Hopefully interesting.

Thanks!

PR #185, opened 2017-10-16
Modified to work with classic or extended tweet format ('text' or 'fu…

lotsoaxolotls:
…ll_text'). This change was applied to the util script wordcloud.py.


edsu:
Thanks!

PR #187, opened 2017-10-26
Adds deletes_api.py.

justinlittman:
Modified deletes.py to use Twitter API. Reasons for this:
* Can lookup based on user id, making it possible to distinguish between accounts that were deleted and accounts that changed screen name.
* Predictability/reliability of API
* Clear expectations of rate limits

### DIFF
--- a/utils/deletes_api.py
+++ b/utils/deletes_api.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+
+"""
+This program assumes that you are feeding it tweet JSON data for tweets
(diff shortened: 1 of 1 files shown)


edsu:
Hey, nice! I didn't realize you could get the same granularity of error messages from the API. If the functionality is the same could this be folded into deletes.py proper instead of having two different versions.

justinlittman:
Definitely. As I'm sure you can tell, I just copied deletes.py and made modifications/enhancements. Logic is the same.

PR #188, opened 2017-10-30
Enhancements to json2csv.py. 

justinlittman:
Added support for splitting into multiple files and including extra fields. Added parsed_date and tweet type field. Added fields related to quote tweets.

Use case for extra fields: I enhanced my tweets with a delete reason field and want to be able to output to CSV.


edsu:
Awesome changes, thanks @justinlittman !

Issue #172, opened 2017-08-21
CSV output


edsu:
When outputting tweets it would be useful to have the CSV option in the command itself. Something like:

    twarc search blacklivesmatter --output csv > tweets.csv

The downside is that this would result in data loss since not everything makes it into the CSV. But perhaps this is something some researchers are willing to live with? At the moment it is a bit clunky constantly referring people to [json2csv.py](https://github.com/DocNow/twarc/blob/master/utils/json2csv.py) but perhaps this isn't too much of a hardship?

If anyone has an opinion I'd love to hear it.

alibama:
this is a great idea - would love to have the option

PR #189, opened 2017-11-21
Made utils a package so that can be re-used.

justinlittman:


justinlittman:
In my case, I want to re-use json2csv.

edsu:
Hmm, I'm kind of reluctant to bring the whole kitchen sink into twarc itself. The idea (which I never properly expressed) was to treat it as a laboratory for ideas that could be pulled into twarc proper. This happened with replies for example.

#172 is about adding CSV output to twarc. This would necessarily bring csv2json into the twarc module in some form. If I prioritized doing this would it satisfy your use case, or were you wanting more of utils to be available upon install?

justinlittman:
That would work for me.

edsu:
@justinlittman when you have a chance can you take a look at what's on master, specifically `twarc.json2csv` now and see if it works for you?

justinlittman:
Looking.

edsu:
This is no longer needed since CSV functionality has been made part of the twarc module.

PR #192, opened 2017-11-28
Update README.md

ppival:
added blurb on how to upgrade


edsu:
`pip install --upgrade twarc` is probably easier since it does not require you to know the version.

PR #199, opened 2017-12-18
Added support for list members.

justinlittman:
This is for retrieving the members of a list by providing the URL of the list.

My use case is extracting the screen names from lists to use as seeds for collecting.


justinlittman:
Here's the link to the API docs: https://developer.twitter.com/en/docs/accounts-and-users/create-manage-lists/api-reference/get-lists-members

PR #200, opened 2017-12-19
Fix NameError: name 'MissingKeys' is not defined

leezu:
The NameError was caught by the program and a infinite loop was entered.


edsu:
Thanks for these fixes -- they are much appreciated.

PR #202, opened 2018-01-04
Handled blocked tweets (401).

justinlittman:
For some unknown reason, I'm blocked by a user. When trying to retrieve a tweet from that user (tweet id 793195186201432064), it results in a 401.

This adds handling for blocked tweets.

### DIFF
--- a/utils/deletes.py
+++ b/utils/deletes.py
@@ -23,8 +23,11 @@ USER_PROTECTED = "USER_PROTECTED"
 USER_SUSPENDED = "USER_SUSPENDED"
 TWEET_OK = "TWEET_OK"
 TWEET_DELETED = "TWEET_DELETED"
+# You have been blocked by the user.
+TWEET_BLOCKED = "TWEET_BLOCKED"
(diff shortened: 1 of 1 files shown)


utils/deletes.py
```diff
@@ -63,7 +66,7 @@ def examine(tweet):
 
         if tweet_status == TWEET_OK:
             return TWEET_OK
-        elif retweet is None:
+        elif retweet is None or tweet_status == TWEET_BLOCKED:
```
edsu: I'm having trouble understanding the logic of what's going on here.

justinlittman: Confusing to try to explain in terms of original, retweet, etc., so let's try this:
If X is a retweet of Y. If X is blocked, then report as blocked.
Otherwise, if X is not blocked then determine status based on Y.

Open to suggestions about how to make the code clearer.

edsu: Yeah, I'm still having trouble understanding this. But since you are actively using this I'll merge it and we can improve it over time as needed.

PR #208, opened 2018-02-21
Handle multiline user descriptions.

justinlittman:
Multiline user descriptions result in multiple lines in the CSV, rather than a single line.

To reproduce: `echo "921446951727652864" | twarc --format csv hydrate -`

### DIFF
--- a/twarc/json2csv.py
+++ b/twarc/json2csv.py
@@ -64,7 +64,7 @@ def get_row(t, excel=False):
       get('created_at'),
       date_parse(get('created_at')),
       user('screen_name'),
-      text(t) if not excel else tweet_url(t),
+      text(t) if not excel else clean_str(text(t)),
       tweet_type(t),
       coordinates(t),
       hashtags(t),
--- a/twarc/json2csv.py
+++ b/twarc/json2csv.py
@@ -55,7 +55,7 @@ def get_headings():
     ]
 
 
-def get_row(t):
+def get_row(t, excel=False):
     get = t.get
     user = t.get('user').get
     return [
@@ -64,7 +64,7 @@ def get_row(t):
       get('created_at'),
       date_parse(get('created_at')),
       user('screen_name'),
-      text(t),
+      text(t) if not excel else tweet_url(t),
       tweet_type(t),
       coordinates(t),
       hashtags(t),
@@ -85,13 +85,13 @@ def get_row(t):
       user('id_str'),
       user('created_at'),
       user('default_profile_image'),
-      user('description'),
+      user('description') if not excel else clean_str(user('description')),
       user('favourites_count'),
       user('followers_count'),
       user('friends_count'),
       user('listed_count'),
-      user('location'),
-      user('name'),
+      user('location') if not excel else clean_str(user('location')),
+      user('name') if not excel else clean_str(user('name')),
       user('statuses_count'),
       user('time_zone'),
       user_urls(t),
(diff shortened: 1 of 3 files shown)


edsu:
I believe that is valid CSV though, and the new lines are kind of significant to the text of the tweet.

justinlittman:
So there's valid CSV and then there's Excel CSV. For Excel, if just open the CSV then line feeds are handled correctly, but tweet and user ids are rounded. If you open with Data > Get External Data and set the id columns to Text, they are not rounded, but line feeds are not handled correctly.

So for Excel, to preserve the ids, line feeds must be sacrificed. Unless I'm missing something, as I am hardly an Excel expert.

edsu:
I'm reluctant to introduce these little tweaks to accommodate Excel. If compatibility with Excel is super important I think I'd be much happier adding an `--output xslx` option that would bring in [xlsxwriter](https://xlsxwriter.readthedocs.io/#) or something similar as a dependency.

edsu:
I guess this is somewhat pointless, but I was curious:

https://twitter.com/edsu/status/966622726017503232

justinlittman:
Excel (sigh) is critical for me, so willing to take whichever approach to make it work.

edsu:
Along with the poll responders I'm definitely leaning towards an xslx output option instead of modifying the text of the tweet. This way users won't have to open the data in a special way, and the actual text of the tweet won't have to be compromised.

jackiekazil:
I am going to chime on this, because this is a topic I feel strongly about. You can't cater to a proprietary format. If Google sheets had its own format, would you also do that one? As someone who has ingested a lot of data, I have never seen anyone who has offered more than a standard CSV. If you think that this is important to your specific users, then I would do it, but only if it has been validated that way. Also, it is important to your specific users at the cost of something else you could be providing and supporting as opposed to this. 

Lastly... *waves* hey!

ianmilligan1:
Long time user of twarc here too – this is an important conversation to have. As somebody who's run a few web archiving workshops and has taken students through twarc a few times, being able to open up CSVs in Excel is key. I would vote for the additional flag, given the ubiquity of MS Office.

I fear if people open up broken CSVs in Excel (despite any disclaimers you might put up about not using Excel) they'll get frustrated and stop using the tool.

justinlittman:
I'm not opposed to XLSX output. (We offer CSV and XLSX from Social Feed Manager.) Single click import into Excel is definitely helpful.

However, one pipeline I've increasingly been using is twarc -> csvkit -> Excel. To support this, I'd propose adding a csv-excel format (perhaps in addition to XLSX), where the newlines are removed.

justinlittman:
Hey @edsu -- Did you reach a verdict on how you'd like to handle this?

edsu:
Based on the poll results I'm not willing to let Excel pollute `--output csv`. Excel simply isn't the only option for looking at spreadsheets anymore, and it doesn't deserve to be treated that way.

Ideally I think twarc would have an `--output excel` which would write out an xslx that could be opened without fuss in Excel.

If you want to add an `--output csv-excel` I won't stand in your way. I'm a bit confused why users need the pristine tweet id in the spreadsheet when they have tweet URL (that contains the tweet id) and also already have most of the data. So that's why I'm not overly enthusiastic about the idea.

justinlittman:
Trust me -- no one is more frustrated with Excel than me. While there are other options, unfortunately Excel is the one option that is consistently on the laptops of the students/faculty I work with.

I'll update this PR so that it provides an `--output csv-excel` option. Thanks for taking the temperature on this one.

justinlittman:
@edsu ready for review

twarc/json2csv.py
```diff
@@ -64,7 +64,7 @@ def get_row(t):
       get('created_at'),
       date_parse(get('created_at')),
       user('screen_name'),
-      text(t),
+      text(t) if not excel else tweet_url(t),
```
edsu: Did you mean to have `clean_str(text(t))` instead of `tweet_url(t)` here?

edsu:
Thanks @justinlittman. I just released as v1.4.0

PR #212, opened 2018-03-25
Add python_requires to help pip

hugovk:
When old Python versions are dropped, this will help pip install the right version for people still running those old Python versions.
 
For more info on how this works:
* https://hackernoon.com/phasing-out-python-runtimes-gracefully-956f112f33c4
* https://github.com/pypa/python-packaging-user-guide/issues/450


edsu:
This was just released as v1.4.1

PR #216, opened 2018-04-05
we don't have a code of conduct

kayiwa:
this is borrowed and attributed to the contributers covenant. It may be
better if @edsu is enforcer? I added myself otherwise.


edsu:
Will referencing https://github.com/DocNow/code-of-conduct not work?

PR #218, opened 2018-04-13
Added filter_users.py script.

justinlittman:
I often find myself needing to filter a collection of tweets by a list of users (screen names or user ids), where that list is too long for convenient jq scripting. This script does that.


edsu:
Thanks @justinlittman !

PR #225, opened 2018-05-16
Show README.md on PyPI

hugovk:
Now the new PyPI (Warehouse) has launched, Markdown is finally supported as the long description for https://pypi.org/project/twarc/.

So instead of https://pypi.org/project/twarc/ showing:

image

you can have the README.md in there.

Make sure these tools are up-to-date:

```bash
pip install -U setuptools twine wheel
```

For more info:
* https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi
* https://github.com/di/markdown-description-example

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,9 @@ from setuptools import setup
 # Also in twarc.py
 __version__ = '1.4.5'
 
+with open("README.md") as f:
+    long_description = f.read()
+
 if sys.version_info[0] < 3:
     dependencies = open(join('requirements', 'python2.txt')).read().split()
 else:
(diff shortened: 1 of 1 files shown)


edsu:
Nice! What are twine and wheel for?

hugovk:
Twine is a tool for uploading to PyPI, and you must use it for this. See step 5 in https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi.

See https://github.com/pylast/pylast/blob/master/RELEASING.md for an example of how I use it.

You can ignore wheel for this PR, I copied and pasted the PR text from another. 

Wheels are a new way to distribute Python packages. You can distribute as both .tar.gz and as a wheel. The big advantage of wheels is when a package contains C code and it can all be precompiled so the user does not need to worry about that. This is especially useful for Windows users where it can be a big pain to set up compilers. And it's a lot quicker.

See https://pythonwheels.com/ for more benefits and further info.

edsu:
Thanks for those details, and for tracking them for twarc!

edsu:
Seems to be [working](https://pypi.org/project/twarc/), although the test suite is failing at the moment :-)

hugovk:
Great! 

Test failure looks unrelated :)

PR #228, opened 2018-06-13
Stream in and out noretweets.

ruebot:
Previous method held everything in memory, which makes working with large datasets difficult. This method should just stream them in and out.


edsu:
👍

PR #230, opened 2018-06-15
Fix to string cleaning in json2csv for python 2.

justinlittman:
See https://github.com/gwu-libraries/sfm-ui/issues/935.

Steps to reproduce:
python2 utils/json2csv.py -x some_tweets.jsonl

Expected result:
Text field is populated with text of the tweet.

Actual result:
Text field is blank.

### DIFF
--- a/twarc/json2csv.py
+++ b/twarc/json2csv.py
@@ -3,6 +3,7 @@
 import sys
 
 from dateutil.parser import parse as date_parse
+from six import string_types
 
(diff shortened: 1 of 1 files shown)


edsu:
Nice catch!

justinlittman:
Unfortunately, caught by an SFM user. Adding more urgency to get SFM off
python 2.

edsu:
This was just uploaded to PyPI as part of v1.4.7

PR #234, opened 2018-07-24
Allow sample/filter to optionally return keepalive events

SamHames:
The keepalive signals returned by the Twitter API streaming
endpoints are useful data in and of themselves. Recording them
allows us to differentiate between quiet periods where no tweets
occurred, and periods where the connection to the API was down.

The format for the keep alive events is just a string message
'keep-alive', and should be easy to consume in downstream apps.
The feature is off by default.

This doesn't solve #119 completely, but does enable the events to
at least be recorded by users of the API.

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -119,6 +119,16 @@ def test_track():
     # reconnect to close streaming connection for other tests
     T.connect()
 
+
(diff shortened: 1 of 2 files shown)


edsu:
Thanks for this!

PR #236, opened 2018-07-26
Bugfix for numeric user_id passed to timeline

### LINKED ISSUE #235: Passing a numeric user_id to Twarc.timeline yields no tweets
This is a minor API usability thing that tripped me up: if you pass the user_id to `Twarc.timeline` as an integer instead of a string, no tweets will be yielded, but the logs show all of the network calls to successfully page through the timeline. 

```>>> import twarc
>>> client = twarc.Twarc(*creds)
>>> len(list(client.timeline(user_id=87818409)))
0
>>> len(list(client.timeline(user_id='87818409')))
3235
```

I think this is because of the comparison on [Line 146 of client.py](https://github.com/DocNow/twarc/blob/master/twarc/client.py#L146) that tries to compare the passed in integer user_id to the "id_str" of the tweet object.

That seems to be the only method that needs that particular pattern, so it could be fixed by converting the integer to a string, but I get the feeling that twarc API expects strings everywhere. I'm happy to submit a fix (whether for that method or documentation).

PS - I have to say it's been really nice to find this library, it's a very approachable codebase, and it's nice to see the focus in the design of the core API. :)

SamHames:
This is the minimal fix, I couldn't find anywhere else this pattern is used so I just took the approach of always converting the user_id to a string for the checks. Hopefully this will make it less surprising for others.

Fixes #235

### DIFF
--- a/twarc/client.py
+++ b/twarc/client.py
@@ -111,6 +111,12 @@ class Twarc(object):
         by the user indicated by the user_id or screen_name parameter.
         Provide a user_id or screen_name.
         """
+
+        if user_id and screen_name:
+            raise ValueError('only user_id or screen_name may be passed')
+        elif not (user_id or screen_name):
+            raise ValueError('one of user_id or screen_name must be passed')
+
         # Strip if screen_name is prefixed with '@'
         if screen_name:
             screen_name = screen_name.lstrip('@')
--- a/twarc/client.py
+++ b/twarc/client.py
@@ -114,7 +114,7 @@ class Twarc(object):
         # Strip if screen_name is prefixed with '@'
         if screen_name:
             screen_name = screen_name.lstrip('@')
-        id = screen_name or user_id
+        id = screen_name or str(user_id)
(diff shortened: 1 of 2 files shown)


twarc/client.py
```diff
@@ -143,8 +143,8 @@ def timeline(self, user_id=None, screen_name=None, max_id=None,
             for status in statuses:
                 # If you request an invalid user_id, you may still get
                 # results so need to check.
-                if not user_id or user_id == status.get("user",
-                                                        {}).get("id_str"):
+                if not user_id or id == status.get("user",
```
edsu: I haven't really traced the logic, but was this a significant change from `user_id` to `id`?

SamHames: I just added the validation for screen_name/user_id arguments, since this PR is right there anyway :)

SamHames:
@edsu Is there anything else you need for this?

edsu:
Whoops, thanks for the bump. This looks great thanks!

Issue #233, opened 2018-07-23
Issue with filter_users.py


kkk279:
Could you please look at the above utility or/and provide better documentation. When it runs, it shows it processed a set number of file but then there isn't an out put. I used it like this, ./utils/filter_users.py  "list file"  main file  > output.  Also not sure of the format of the "list file". Thank You.

edsu:
@justinlittman any ideas?

justinlittman:
The format for the list file is documented in the comments. See https://github.com/DocNow/twarc/blob/master/utils/filter_users.py#L5

Happy to try to troubleshoot. What OS, python version are you using? Can you provide a sample list file and main file so I can try to reproduce?

kkk279:
Hi, I am using ubuntu and python 2.7. The main file is a 40+ gig twitter json file and I have attached  a sample(s) of the csv list in txt format since this thread doesn't support CSV.
[Username.txt](https://github.com/DocNow/twarc/files/2220305/Username.txt)

file.

kkk279:
I created some sample files so you can troubleshoot if need be.

[cajun_navy.zip](https://github.com/DocNow/twarc/files/2221147/cajun_navy.zip)
[cajun_scrnname.zip](https://github.com/DocNow/twarc/files/2221148/cajun_scrnname.zip)

This is the code I used,
python ./utils/filter_users.py cajun_scrnname.csv cajun_navy.json
 > output.json

This is the result:  2018-07-23 16:42:54,236 INFO processed 0 tweets
With no output file.  When I use the larger file, it processes 1,000,000 tweets but does not spew out any result. Thank You for your help.

justinlittman:
Ahh, it should be one screen name per line like:
```
foo
bar
```

Let me know if that solves the problem. If so, I'll clarify in the docs.

kkk279:
Hi,
That didn't work. I tried that just now using both txt and csv as list files. It does print out an output in both Json and CSV after processing but its always 0 KB in size.

python ./utils/filter_users.py usernames.csv ./data/*.json > output.json

justinlittman:
I didn't get the earlier tweet file. Can you email it to me directly? (justinlittman@gmail.com)

kkk279:
Thank You. Just did.

PR #237, opened 2018-07-26
refs #233. Handles carriage returns in user list file for filter_user…

justinlittman:
…s.py. Also, clarifies format of user list file.


edsu:
Thanks @justinlittman

PR #240, opened 2018-08-08
Pagination enhancement and optimisation

SamHames:
Background: I have a use case that requires polling a lot of not very active timelines, and for this case the need to retrieve an empty page of results to indicate the end of pagination, and the need to page through the entire set of results can have a big impact on throughput.

This PR adds to `Twarc.timeline` and `Twarc.search`:
- An optimisation around since_id handling that in the happy path lets us skip an API call that would not return any more results - this change should be transparent to the user, it just means less API calls are made. This works by making the since_id in the call inclusive, and using the tweet with that since_id as a sentinel for the end of pagination.
- A (default off) argument max_pages that limits the number of pages retrieved instead of retrieving everything.

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -45,7 +45,9 @@ def test_search():
 
 def test_search_max_pages():
     tweets = list(T.search('obama', max_pages=1))
-    assert len(tweets) == 100
+    assert 0 < len(tweets) <= 100
+    tweets = list(T.search('obama', max_pages=2))
+    assert 100 < len(tweets) <= 200
 
 
 def test_since_id():
@@ -214,6 +216,9 @@ def test_timeline_max_pages():
 
     first_page = list(T.timeline(user_id=user_id, max_pages=1))
     assert 0 < len(first_page) <= 200
+
+    all_pages = list(T.timeline(user_id=user_id))
+    assert len(all_pages) > len(first_page)
 
 
 def test_timeline_by_screen_name():
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -43,6 +43,11 @@ def test_search():
     assert count == 10
 
 
+def test_search_max_pages():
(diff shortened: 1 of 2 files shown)


edsu:
I'm a bit confused about why `max_pages` is needed. If I want to do a search and I'm only interested in a certain number of results why can't I:

```python

count = 0
for tweet in t.search('obama'):
    print(tweet['id'])
    count += 1
    if count > 100:
        break
```

Is it just a convenience for not having to count things yourself?

SamHames:
In the case of search, yes, it's just a convenience because the pagination is consistent (and that is a very simple approach that I'm embarassed didn't occur to me first :) ). I was more interested in user timelines, I added it to search as well because the control flow and the method signatures are more or less identical so I don't think it hurts to keep them consistent.

For user timelines that approach is a little bit messier because the pages aren't necessarily consistent sizes because of how deleted/suspended tweets are removed from the page after the count is retrieved: [`The value of count is best thought of as a limit to the number of Tweets to return because suspended or deleted content is removed after the count has been applied`](https://developer.twitter.com/en/docs/tweets/timelines/api-reference/get-statuses-user_timeline.html). 

The applications I have in mind are more about controlled sampling in the face of rate limits, and making the most of every API call: if I have retrieved tweets from the API I want to grab all of them, but I don't want to make an extra API call to reach an arbitrary number of tweets.

I'm happy to remove the max_pages if you're not sure about how well it fits in with Twarc: I can work around it, but it's a little messy, and I want to build some applications on top of twarc so it's good for me to know your thinking about the direction of the library so I'm not working against it. (I think this is also the last little niggle I have as well, it's been a pleasure to work with the codebase).

edsu:
Ok, this makes sense, thanks for taking the time to explain. Let's roll with it and see how it works in practice!

PR #241, opened 2018-08-10
Update README.md to include installation of pip

skilfullycurled:
In the process of directing someone who had little experience using the terminal, I realized this was also required (at least on OSX).  So I figured, as long as we're starting the instructions at the installation of python itself, the installation of pip might be a good addition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ generate an access token and access token secret. With these four variables
 in hand you are ready to start using twarc.
 
 1. install [Python](http://python.org/download) (2 or 3)
-2. pip install twarc (if upgrading: pip install --upgrade twarc)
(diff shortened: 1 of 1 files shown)


edsu:
This assumes that easy_install is installed which is not always the case either. pip comes with Python when you install it, but isn't included by Apple on OS X. I would prefer to link to the pip installation instructions: https://pip.pypa.io/en/stable/installing/

skilfullycurled:
Ahhhhhhhh.  I thought easy_install was preinstalled on OSX and I didn't realize that pip was included in a "manual" python installation.  So that really changes the equation of reasoning about ease of use.  Thanks, Ed!

Issue #227, opened 2018-06-03
A better authentication error message


edsu:
There should be a better error message when a set of keys don't work:

```
Traceback (most recent call last):
  File "/usr/local/bin/twarc", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/twarc/command.py", line 231, in main
    for thing in things:
  File "/usr/local/lib/python3.6/site-packages/twarc/client.py", line 213, in follower_ids
    raise e
  File "/usr/local/lib/python3.6/site-packages/twarc/client.py", line 209, in follower_ids
    resp = self.get(url, params=params, allow_404=True)
  File "/usr/local/lib/python3.6/site-packages/twarc/decorators.py", line 36, in new_f
    resp.raise_for_status()
  File "/usr/local/lib/python3.6/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Authorization Required for url: https://api.twitter.com/1.1/followers/ids.json?screen_name=anarchivist&cursor=-1&tweet_mode=extended
```

SamHames:
I've run into this in a few places, mainly attempting to retrieve tweets for now protected accounts. 

Am I right in my thinking that the fix for this would be to catch and raise a more specific error in the `rate_limit` decorator on a 401 response? One tricky part might be that the 401 could be because the credentials are bad, or just don't have access to that specific user if they're protected.

edsu:
Yes, I think the difference between these two types of errrors is important for the user: invalid credentials vs lack of permission. I think the text of error message itself could be used to determine which one happened?

SamHames:
That's what I was hoping, I'll do some investigation and submit a PR sometime in the next couple of days.

Another option is to use the [`account/verify_credentials`](https://developer.twitter.com/en/docs/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials.html) endpoint and hook that into the `check_keys` method. That might be a more little complex, but it would catch the invalid credentials case earlier and allow raising a specific error to the user rather than trying to parse an error message to disambiguate.

PR #242, opened 2018-08-14
Catch and display better errors on authentication failures (401)

SamHames:
Based on the discussion in #227:

1. Catches and raises an error with a more specific message when a method hits an authentication failure (HTTP 401), such as for protected accounts or invalid credentials. 
2. To better handle the issue of invalid credentials from the beginning, adds key validation that explicitly checks against the Twitter API on client initialisation. This makes it easier to flag when the credentials are invalid and explicitly notify the user.

One issue I want to flag is that key validation is currently on by default, but that endpoint only allows 75 requests/15 minutes. I'm not really sure how people are making use of the CLI, so that might be a little restrictive with lots of invocations.

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -539,3 +539,17 @@ def test_extended_compat():
 
     assert 'full_text' in next(T.timeline(screen_name="BarackObama"))
     assert 'text' in next(t_compat.timeline(screen_name="BarackObama"))
+
+
+def test_invalid_credentials():
+    old_consumer_key = T.consumer_key
+    T.consumer_key = None
+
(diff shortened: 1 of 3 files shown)


edsu:
This is a awesome @SamHames! I like the idea of checking the keys when the client is created a lot. I think if people are constantly recreating the client when fetching data they are doing it wrong, since many of the retrieval methods work as generators. One thing I considered doing is introducing an TwarcException which could be raised & caught in order to just print the error message without the full distracting stack trace when on the command line. But I think this is perfect for now.

PR #248, opened 2018-08-29
Skip key validation for the configure command

SamHames:
Also add a CLI argument to skip key checking for good measure.


edsu:
@SamHames thanks for the quick fix on this one -- and for keeping a better eye on Slack than me :-)

PR #251, opened 2018-09-17
we can install twarc via homebrew on macOS

kayiwa:
cannot decide if this is simpler or not but pleasantly surprised to find
a keg of twarc

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ in hand you are ready to start using twarc.
 1. install [Python](http://python.org/download) (2 or 3)
 2. pip install twarc (if upgrading: pip install --upgrade twarc)
 
+### Homebrew (macOS only)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,14 @@ in hand you are ready to start using twarc.
 1. install [Python](http://python.org/download) (2 or 3)
 2. pip install twarc (if upgrading: pip install --upgrade twarc)
 
+### Homebrew (macOS only)
+
+For macOS users, you can install `twarc` via:
+
+```bash
+$ brew install twarc
```
hugovk: Tip: use bash when there's no prompt and console when there is a prompt.

For example, bash with no prompt:

```bash
brew install twarc && example1
example2
```

console with prompt:

```console
$ brew install twarc && example1
here's some output
$ example2
more output
```

PR #252, opened 2018-09-17
Fix for issues with creating dedup.db inside the archive directory

Segerberg:
The script initated two dedup.db one where you ran the script and one inside the archive directory. 
This PR fixes that issue.


edsu:
Whoops, thanks!

PR #262, opened 2019-01-15
Changed network graph output to directed graphs

decapstrike:
Since Twitter allows for one directional follows, and retweets and replies are inherently directional, a directed graph is preferable to an undirected graph. This change simply changes the default to directed. Next an option will be added to select undirected or directed.

### DIFF
--- a/utils/network.py
+++ b/utils/network.py
@@ -67,7 +67,7 @@ if len(args) != 2:
 
 tweets, output = args
 
-G = networkx.Graph()
+G = networkx.DiGraph()
 
 def add(from_user, from_id, to_user, to_id, type):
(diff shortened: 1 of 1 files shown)


decapstrike:
Absolutely, this tool is fantastic so I want to help anyway I can. Do the utils get installed when using the PIP install method?

edsu:
Alas no, I think there are some undocumented dependencies in some of them. I would to move towards doing that though. Perhaps using a sumbcommand style:

    % twarc network tweets.jsonl tweets.html

which would call `twarc-network` that gets installed? I think they would need to be prefixed if going into a system path.

PR #270, opened 2019-03-08
improve error handling with additional exceptions

flaviomartins:
Handle a few additional exceptions that can be raised by 'requests' when handling a request

### DIFF
--- a/twarc/client.py
+++ b/twarc/client.py
@@ -8,6 +8,10 @@ import types
 import logging
 import requests
 
+import ssl
+from requests.exceptions import ConnectionError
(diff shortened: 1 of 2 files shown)


edsu:
Thanks! Did you actual trigger these exceptions when doing data collection with twarc?

flaviomartins:
Yes. I collected the exceptions I hit while using the hydrate command. I believe the traceback was one of those "an exception was thrown while handling an exception".

edsu:
Ok, I've merged this but I'm a little bit concerned about the ssl dependency that is now in twarc. Can you tell me what version of python and the requests module you were using when you ran into the additional errors?

PR #272, opened 2019-03-11
Moves to using a named logger.

### LINKED ISSUE #271: Switch to using a named logger instead of root logger
When using Twarc as a library, filtering out Twarc logging messages is problematic because Twarc logs as the root logger. Switching to a named logger, e.g., (_twarc_) would help.

Wanted to get feedback before PRing.

justinlittman:
Also, changed deprecated `.warn` to `.warning`.

This change can be tested by adding / removing `logging.getLogger("twarc").setLevel(logging.ERROR)` after basic config `logging.basicConfig` in `commands.py`.

closes #271

### DIFF
--- a/twarc/client.py
+++ b/twarc/client.py
@@ -29,6 +29,8 @@ else:
     import configparser
     from urllib.parse import parse_qs
 
+log = logging.getLogger('twarc')
+
 
 class Twarc(object):
     """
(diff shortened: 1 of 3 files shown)


edsu:
This looks great, thanks @justinlittman!

edsu:
This was just released to PyPI as v1.6.3

PR #276, opened 2019-04-20
Add handling of protected accounts

dshahrokhian:
The script now does not crash upon finding a protected account.


dshahrokhian:
There seems to be some problems with the credentials, can you check the issue and re-run the test?

edsu:
Thanks for this!

edsu:
This was released as part of v.1.6.5

PR #277, opened 2019-04-22
README: fix links to Wikipedia

guyfawcus:
A while age, a hyphen was added to the "Line-delimited JSON" section
of the "JSON streaming" Wikipedia page that the README links to.
(https://en.wikipedia.org/w/index.php?title=JSON_streaming&type=revision&diff=prev&oldid=815227476)

This updates the links in all of the README files to use hyphens.


edsu:
Thank you!

PR #279, opened 2019-04-24
Count all emoji in utils script

### LINKED ISSUE #278: emoji.py does not capture all emoji
Currently, the emoji counter script in utils/emoji.py does not actually capture all emoji, despite its name. It only looks for the Unicode range `\U0001f600-\U0001f650`, which corresponds to emoticons/smileys, but not other types of emoji.

frederik-elwert:
This pull request is an attempt to fix #278.

I renamed the old script emoticons.py, because that better reflects what it does. I added a second script emojis.py to count all emoji (based on the external emoji module).


edsu:
Why don't we delete emoticons? Or does it not report a subset of what emojis.py will?

hugovk:
I thought emoticons are things like `:)` `:-D`

frederik-elwert:
> Why don't we delete emoticons.py? Does it not report a subset of what emojis.py will?

Yes, we could also do that. I was a bit hesitant, because a) it runs significantly faster on large collections, b) it does not have external dependencies, and c) the emoji set also covers things like arrows and other symbols that people might find irrelevant. But since the unicode range in the regex will most probably also leave some things out that people would expect to be included, I’m really fine with removing it.

> I thought emoticons are things like `:)` `:-D`

The use of the term is not very consistent. As I understand it, historically it refers to the ASCII form, but now also includes their pictorial representations including unicode characters. I was mainly referring to the Unicode block of the same name, because that’s what the old script captures.

edsu:
Awesome, thanks!

edsu:
If you do end up doing any reporting with it and are comfortable with sharing please let me know on twitter.com/edsu

Issue #284, opened 2019-05-24
twarc-archive.py uses tweet_mode compat as default


frederik-elwert:
The script `twarc-archive.py` sets `tweet_mode` to `compat` by default. This is unexpected, as in all other places, twarc uses `extended` as the default mode. Users might not be aware of this behaviour, which can cause in incomplete data.

Since tools like json2csv.py handle the absence of `full_text` silently and fall back to the truncated `text`, users are not made aware of the fact that the collected tweets contain truncated text.

The default for `tweet_mode` should be changed to `extended`, matching twarc’s other commands.

edsu:
Thanks for noticing this and sending the PR. It has been merged!

PR #286, opened 2019-05-24
Change default `tweet_mode` to `extended`

frederik-elwert:
Change default `tweet_mode` to match the default in other places. See issue #284.


edsu:
Thanks for noticing this, and taking the time to send a PR. Much appreciated!

PR #287, opened 2019-05-29
max_pages parameter in friend_ids and follower_ids

FlxVctr:
As some users follow/are followed by millions of people, only collecting the most recent ones, comes in handy at times. If you agree, I'm happy to test these proposed code changes. I apologise for not having found any contributor's guidelines. Please let me know if I've overlooked them.

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -28,6 +28,7 @@ You will need to have these environment variables set to run these tests:
 logging.basicConfig(filename="test.log", level=logging.INFO)
 T = twarc.Twarc()
 
+
 def test_version():
     import setup
     assert setup.__version__ == twarc.__version__
@@ -117,6 +118,7 @@ def test_geocode():
 
     assert found
 
+
 def test_track():
     tweet = next(T.filter(track="obama"))
     json_str = json.dumps(tweet)
@@ -191,6 +193,7 @@ def test_locations():
     # reconnect to close streaming connection for other tests
     T.connect()
 
+
 def test_languages():
     count = 0
     ok = True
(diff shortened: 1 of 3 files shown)


edsu:
Thanks for this. The reason why twarc returns a generator rather than a list is that 1) they could be pretty long lists and 2) it gives control over the iteration back to the user of twarc. So you can do something like this to keep a count and decide when to stop:

```
import twarc

t = twarc.Twarc()

count = 0
for follower_id in t.follower_ids('barackobama'):
    print(follower_id)
    count += 1
    if count > 2000:
        break
```

I would prefer to keep things this way rather than introducing other controls like the one you are suggesting. Unless there is some problem with this approach?

FlxVctr:
One problem with the approach occurs if you want to control how many API calls you spend. As pages might have less than the possible maximum of account ids, one might waste an API call for almost nothing.

I thought the max_pages parameter would be in line with what's already implemented for other API endpoints (e.g. user timeline).

edsu:
Ahh, you're right -- I forgot that max_pages was on search and timeline methods already. This was a slippery slope :-) If you add a unit test I'll merge it in the interests of consistency.

FlxVctr:
Unit tests done and passing for Python 3. Did not test for Python 2 but don't see why it should be different there.

Not sure why the version in __init__.py was not matching for the version test that was failing. If you had good reasons to keep it lower, just drop the latest commit … or let me know and I'll do it.

Cheers :)

edsu:
Thanks @FlxVctr!

edsu:
This was just released as v1.7.3

FlxVctr:
Great! Thanks for the acknowledgements in the release and the seamless contribution and review process :)

Issue #290, opened 2019-07-09
--profile argument on twarc-archive.py does nothing


slarrain:
The `twarc-archive.py` utility has an option for `--profile` where one can specify the profile to use from the `.twarc` config file.
https://github.com/DocNow/twarc/blob/22106ea25710e71e166e68ebd22586e9863150eb/utils/twarc-archive.py#L65

However, looking at the code further down, that argument is never used! 
https://github.com/DocNow/twarc/blob/master/utils/twarc-archive.py#L101-L106

Is there a bug or am I missing something?

slarrain:
Add a PR that should solve this issue. #291

slarrain:
PR #291 got merged, so closing this issue.

PR #291, opened 2019-07-09
Add profile when creating Twarc on twarc_archive

slarrain:
This script had the --profile argument but did not implement it
when creating the Twarc instance.
This PR adds it, closing #290


edsu:
Whoops, thank you @slarrain! I think this is one more reason why the archive functionality should be merged into the twarc utility proper instead of being a sidecar utility.

slarrain:
I agree @edsu . For me, `twarc_archive.py` is more of my go to tool that regular `twarc`.
Anyway, thanks so much for the super quick Merge and specially for this awesome library.

PR #294, opened 2019-08-26
Add docker and docker compose files

breyten:
I like docker ;) This is a simple setup so that you can use twarc on the command line in a docker container.


edsu:
I like Docker too! But I feel like twarc could use a simpler Docker expression. I'm closing this for now, but please reopen if you think it can be improved.

PR #299, opened 2019-10-26
Fix typo in reference to `quoted_status_id_str`

freeatnet:
Hi there!

I was experimenting with twarc and noticed it looks for a `quotes_status_id_str` field on a tweet to find related threads. It looks like there is no `quotes_status_id_str` on a tweet, but there is a `quoted_status_id_str` that points to the tweet quoted in the current tweet. Is that what was meant here?


edsu:
Nice catch, thanks @freeatnet!

edsu:
This was just released as part of v1.7.5

PR #300, opened 2019-11-05
add optional arguments for search queries

dshahrokhian:


edsu:
Thanks for this. Can you share any details about the use case for adding this option to return less data?

dshahrokhian:
Sure, for example when crawling tweet ids of a specific user. Or scenarios in which you just care about the tweet object and can save up the space of the user object (when dealing with millions of tweets).

dshahrokhian:
@edsu any updates? I have some other features in the pipeline

edsu:
Sorry this slipped off the radar. Thanks for the friendly prod.

edsu:
It looks like `test_max_and_since_ids` fails now. I'm looking into why.

edsu:
Ah, some of the subtle logic around max_id needed to be restored. 

@dshahrokhian also, while adding a test for the new search parameters `from_date` and `to_date` I see that the `fromDate` and `toDate` don't work with the standard search API. Unless you object I'm going to remove them.

dshahrokhian:
My bad, I did not realize that the parameters only work with the Premium API, for which I do have access. 

Regarding the logic, it was my fault, I thought I had reverted them before performing commit.

dshahrokhian:
I guess that Premium API support is out of the scope of this project (?)

edsu:
Premium API support isn't out of scope. But it would require more changes, e.g. the URL endpoints. If you want you can open an issue for Premium support and we can try to hash out a design for that.

dshahrokhian:
Awesome! I have made a wrapper for premium search for now. Will try opening an issue soon to see what is the best way to integrate this without breaking anything :)

Issue #245, opened 2018-08-20
Use of multiple accounts when rate limited [New feature


slarrain:

slarrain:
I've been thinking of a way to extract multiple tweets and accounts and bypass (sort of) the rate limit imposed by Twitter.

Here is what I thought:

What if we have multiple accounts configured on twarc, and the moment we get rate limited, we go to the next account in our `.twarc` file and so on, until we reached our final account (or the original) and then we actually wait until the timeout is over. 
Obviously, this would be an optional feature for the user and not by default. But I think that, at least in my case, it would be very useful. But I'm not sure if it would be useful for anybody else.

I've even been looking into how to implement it, and I think the best way would be on the decorator function, that actually waits (sleeps) for the timeout.

What do the rest of the community think? Is it worth implementing it?

slarrain:
That is indeed problematic. I have not seen that part of the Twitter Developer Policy and Agreement. Thanks for pointing it out.

edsu:
I think it definitely would be problematic re the TOS to register multiple app keys. I think it also could be weird if multiple accounts are registered by one person as a type of sock puppet to get multiple access keys. One thing I’ve been curious about in the case of [DocNow](https://github.com/docnow/DocNow) is whether a group of actual people could work collaboratively to document an event, and contribute their keys to the effort.

slarrain:
Oh, I was always thinking of multiple keys from different people. Which, reading the TOS again, could actually be acceptable. One person would not be circumventing the rate limit. But instead, multiple people would be using their own keys, which in theory, should be acceptable. And if all of them reach the limit, then they would have to wait.
Is actually a way of having a group of people make the same search together.

edsu:
twarc itself doesn't really offer any functionality for group collaboration -- but this is something we are thinking about enabling in the [docnow](https://github.com/docnow/docnow) application. if you have any thoughts about how that could work, please add them to that issue tracker!

SamHames:
Yeah - I should have been a little bit more nuanced because I think in terms of complying with the terms of service it matters where those keys came from, and how they're being used. 

With multiple people contributing keys (I'm assuming doing the OAuth dance to collect user credentials?) the tricky parts in the terms of service are about how the data is shared. Something I'm trying to get my head around in the research context/institutional ethics frameworks right now.

The other thing to keep in mind is that the recent changes to the developer signup process mean that it's not going to get any easier to get application credentials since Twitter actually seems to be reviewing them more closely now.

[...]

PR #303, opened 2019-11-27
Carousel

slarrain:
After it hits the rate limit of one of the users on the config `.twarc` file, it goes into the next user/credentials. After all have hit the rate limit, it waits for the time of the one that has the shortest wait time. And then continues from that one on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ build
 dist
 twarc.egg-info
 .pytest_cache
+dev_env/
+dev_venv/
--- a/twarc/decorators.py
+++ b/twarc/decorators.py
@@ -26,10 +26,10 @@ def rate_limit(f):
         # Create the neverending pool or carousel
         pool = cycle(useraccounts)
 
-        # If it was note set, then it if the first
+        # If it was note set, then it is the first
         if self.profile == '':
             self.profile = next(pool)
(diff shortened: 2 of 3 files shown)


edsu:
Hi @slarrain -- did I miss something about why you decided to close this?

slarrain:
Hi @edsu !

Sorry for the late reply. 
So, on #245 I raised an issue that I thought could be interesting. Say we have a team of 3 people and we all want to use our credentials to mine, for example, "Trump".
On its current state, we can't do that automatically. When one gets rate limited, it doesn't go automatically into the next credentials. I wanted that behavior, but it was met with some resistance on on the #245 issue.
So I forked Twarc, and implemented it on my fork.
But when I did the PR from the branch to merge into Master, instead of trying to merge with my fork, I accidentally tried to merge into your actual Twarc. So I closed the PR, and did it on my fork and it is now working as expected.

But if you think this behavior is something Twarc could actually use, I'd be more than willing to submit a PR to the actual package (yours), provided that you give me some guidance on the Code Review as to what is missing in terms of code and documentation to actually be "mergeble"

Of course, feel free to take a look around at the PR on my fork and let me know if you think it could work or if it is a functionality that should live on a fork.

Thanks again for this great package.

PR #310, opened 2020-02-06
Create d3_network.py util

eam5:
This is really just an update to the network.py util, that makes a more user friendly d3 display. I made it a separate file, since I'm unfamiliar with how twarc users typically use the network.py for gexf, etc, and I'm not sure my changes if effect that since Im new to python. Please let me know if you incorporate these changes into the network.py instead, so I can update my documentation. Thank you!

### DIFF
--- a/utils/d3_network.py
+++ b/utils/d3_network.py
@@ -1,0 +1,369 @@
+#!/usr/bin/env python
+
+# build a reply, quote, retweet network from a file of tweets and write it
+# out as a gexf, dot, json or  html file. You will need to have networkx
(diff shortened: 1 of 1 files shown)


edsu:
Thanks for this @eam5. The D3 visualization definitely has needed some love, so I appreciate you taking a look at it. Would it be ok if I see how easy it would be to merge your enhancements into network.py?

edsu:
Please reopen this PR or open a new one if you get around to merging your changes into network.py. I don't think we need two d3 network visualizers!

PR #314, opened 2020-03-12
add option for number of emojis

melaniewalsh:
This might not be the most efficient way to revise this script, but I thought it might be useful to add an option for the number of emojis that a user wants to return. If it's a big tweet dataset, there can be tons of emojis, and it seems useful to be able to see just the top n emojis.

### DIFF
--- a/utils/emojis.py
+++ b/utils/emojis.py
@@ -4,14 +4,30 @@ import re
 import json
 import fileinput
 import collections
+import optparse
 
 import emoji
+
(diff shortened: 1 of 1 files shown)


edsu:
Looks good to me!

melaniewalsh:
@edsu If you decide to incorporate this option, I can also add an explanatory note in the code! Just let me know

edsu:
I thought I merged this already, my apologies @melaniewalsh!

PR #315, opened 2020-03-13
include accurate favorite counts for retweeted tweets

melaniewalsh:
When exploring tweet CSV files created with json2csv.py, I noticed that retweeted tweets contain the same "retweet_count" information as the original tweet but they do not contain the same "favorite_count" information. All retweeted tweets show up as having 0 favorites, even though the original tweet might have thousands of favorites.

This update includes accurate favorite counts for retweeted tweets AKA the same favorite count as the original tweet.

### DIFF
--- a/twarc/json2csv.py
+++ b/twarc/json2csv.py
@@ -71,7 +71,7 @@ def get_row(t, excel=False):
       hashtags(t),
       media(t),
       urls(t),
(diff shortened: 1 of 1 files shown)


edsu:
Awesome, thanks @melaniewalsh!

PR #325, opened 2020-04-30
Support Twitter labs v1 sample endpoint

SamHames:
This is a work in progress to start figuring out how to support the new Twitter API as previewed in the labs endpoints. The motivating factor was to figure out how/if we could support [this covid endpoint for researchers](https://developer.twitter.com/en/docs/labs/covid19-stream/overview).

This ignores the problem of how to handle the data once collected - the focus here is just preserving the data as is, so options like CSV support aren't currently considered.

## Issues that are already handled:

1. Authentication - the recent addition of app only authentication covers authentication for many of the new endpoints :)
2. Not surprisingly most of the connection handling and retry logic is still appropriate - thought there will need to be some minor changes (see below).

## Issues to be resolved:

1. Data formats and handling: there are multiple options in the labs endpoint for controlling exactly what is returned. I think the options for twarc are to pick a comprehensive default for what can be archived, or maybe expose a way to configure this easily.

    - in the V1 endpoints there is a format option, which is either `compact`, `default` or `detailed`, with default returning the most information.
    - in the V2 endpoints, the format is replaced with `fields`, where you can specific a detailed list of fields to be included in the tweet.
    - There are also expansions (basically expanding a referenced object such as a quoted tweet or retweeted user directly in the response.)
    - The are also `annotations` which apply some magic to markup tweets with various entities (such as people and places).

2. How to support the current and labs endpoint in parallel and smoothly enable a transition between them. 

    - I think we can add all of the labs endpoints as additional methods, living side by side with existing methods - there will be work required to make sure all of the argument handling is consistent 
    - This also lets us handle v1, v2 and whatever else comes next on an
    endpoint by endpoint basis.

3. I think the twarc client get and post methods probably need a little bit of work to better handle points 1 and 2, and new failure modes:

    - probably need a nicer way to handle different parameters for labs v1 + v2 alongside the existing endpoints
    - need to review any new errors and failure modes - for example some endpoints have a new monthly cap on the number of tweets transferred

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -631,7 +631,7 @@ def test_invalid_credentials():
     T.consumer_key = old_consumer_key
 
 def test_app_auth():
-    ta = twarc.Twarc(app_auth=True) 
+    ta = twarc.Twarc(app_auth=True)
     count = 0
     for tweet in ta.search('obama'):
         assert tweet['id_str']
@@ -639,3 +639,20 @@ def test_app_auth():
         if count == 10:
             break
     assert count == 10
+
+
+@pytest.mark.xfail
+def test_labs_v1_sample():
+    ta = twarc.Twarc(app_auth=True)
+
+    collected = 0
+
+    for tweet in ta.labs_v1_sample():
+        if 'data' in tweet:
+            collected += 1
+
+        if collected == 100:
+            break
+
+    # reconnect to close streaming connection for other tests
+    ta.connect()
--- a/twarc/client.py
+++ b/twarc/client.py
@@ -284,7 +284,7 @@ class Twarc(object):
             for user_id in user_ids['ids']:
                 yield str_type(user_id)
             params['cursor'] = user_ids['next_cursor']
-            
+
             if max_pages is not None and retrieved_pages == max_pages:
                 log.info("reached max follower page limit for %s", params)
                 break
@@ -304,7 +304,7 @@ class Twarc(object):
             params = {'screen_name': user, 'cursor': -1}
 
         retrieved_pages = 0
-        
+
         while params['cursor'] != 0:
             try:
                 resp = self.get(url, params=params, allow_404=True)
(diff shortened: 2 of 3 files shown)


edsu:
Thanks for this well thought out PR @SamHames! Do you think we should continue to work on this branch, or merge what you have so far and add new things in? 

Re: 1 I like the idea of a comprehensive default that pulls everything. I can't think of any downsides to this approach, unless some options preclude others, which doesn't seem to be the case.

SamHames:
For this first pass I'm thinking of adding all the v1 endpoints that aren't already replace by v2 endpoints as a starting point just to see if any other thoughts shake loose - I'm still not quite sure of the best approach for testing yet either. Changing that `get` method makes me nervous about breaking existing functionality :sweat_smile:.

If you're happy with a comprehensive default then round 2 of this work can be to actually figure out what that should be and add support for all of the v2 endpoints.

edsu:
Sorry to have dropped the ball on this @SamHames -- do you think I should merge this in, or is it waiting for additional work?

SamHames:
Ah, it's me dropping the ball here, it's been on my todo list for weeks to get back to this - I'm looking at it right now to try and add the extra v1 endpoints.

I will also do some soak testing with the sample endpoint just to see if anything interesting comes up.

SamHames:
On reconsideration:

- The only other endpoints that are still at v1 are filtered stream, which is closed to new signups for now, and tweet metrics, which I don't think I have a meaningful use case to test for in twarc.

So I'm just going to run the sample stream overnight, and see if anything interesting happens, if that's all good I reckon we can merge this. (I also added a very rough test case, but marked as expected to fail in pytest because it requires the account making the requests having access to the labs endpoints.)

SamHames:
Update: ran an overnight recording test for this, and I don't seem to have broken anything, so I don't think i have anything else to add - open to suggestions if you think there's anything else?

edsu:
Looks good @SamHames thanks for continuing to push on this!

PR #337, opened 2020-07-02
Fix broken README links: master->main

hugovk:


edsu:
Thank you for your attention to detail @hugovk!

PR #346, opened 2020-09-20
English and pt/jp documentation fixes/updates

lxcode:
A few minor corrections in the English README, and adding info on new functionality to the Portuguese version.


edsu:
Thank you for doing all this work @lxcode!

PR #348, opened 2020-09-24
Light linting and cleanup for util scripts

lxcode:
Nothing structural, just removing a couple obsolete references, making permissions uniform, and making language servers happy.


edsu:
Thanks for these @lxcode!

PR #353, opened 2020-10-16
Update unshrtn script to expand user profile URLs

milesmcc:
Sometimes, the user profile URL is shortened (either by Twitter into a `t.co` link or manually by the user). This PR factors the unshortening code out of the `rewrite_line` function in order to apply it to both `tweet["entities"]["urls"]` as well as `tweet["user"]["url"]`.

Open questions:
* Should there be a way to specify which hostnames are unshortened/which are excluded? Currently, `twitter.com` URLs are excluded, but I could imagine someone wanting to only unshorten `t.co` URLs.
* I removed the check that skips unshortening lines that have already been unshortened, as I don't think it was working in the first place (it checked for `tweet["unshortened_url"]`, but the unshortened urls were previously written to `tweet["entities"]["urls"][n]["unshortened_url"]`) and would now also need to check for profile urls. (I can reimplement it if we think it's necessary, but I'm not sure it is given that unshrtn caches.)

(Made with some guidance from @lxcode.)

### DIFF
--- a/utils/unshrtn.py
+++ b/utils/unshrtn.py
@@ -75,10 +75,6 @@ def rewrite_line(line):
         logging.error(e)
         return line
 
-    # don't do the same work again
-    if "unshortened_url" in tweet and tweet["unshortened_url"]:
-        return line
-
     for url_dict in tweet["entities"]["urls"]:
         if "expanded_url" in url_dict:
             url = url_dict["expanded_url"]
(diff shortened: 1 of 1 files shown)


milesmcc:
(Happy to revert the changes in `main` from my autoformatter if they're too opinionated!)

edsu:
Thanks for the TLC to improve this utility!

PR #354, opened 2020-10-19
Add support for max date in `filter_date` utility

milesmcc:
This PR reworks the filter_date utility to add support for a `--maxdate` flag (in addition to just `--mindate`).

While these changes retain backwards compatibility, they use a more sound argument parsing system and allow more flexibility in the order of arguments.


edsu:
Thanks @milesmcc!!

milesmcc:
Thank you!

PR #355, opened 2020-10-22
Support gnip full archive search

epicfaace:
Support full archive search with the gnip twitter enterprise API.

This PR probably still needs some cleanup (and tests), but wanted to get feedback on the authentication mechanism that was needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,21 @@ non-sandboxed environments is 500)
       --sandbox \
       > tweets.jsonl
 
+## Gnip Enterprise API
+
+Twarc supports integration with the Gnip Twitter Full-Archive Enterprise API.
+To do so, you must pass in the `--gnip_auth` argument. Additionally, set the
+`GNIP_USERNAME`, `GNIP_PASSWORD`, and `GNIP_ACCOUNT` environment variables.
+You can then run the following:
+
+    twarc search blacklivesmatter \
+      --gnip_auth \
+      --gnip_fullarchive prod \
+      --from_date 2014-08-04 \
+      --to_date 2015-08-05 \
+      --limit 1000 \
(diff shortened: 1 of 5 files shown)


edsu:
Wow, this is super @epicfaace! I didn't realize GNIP was still a thing, I thought it was replaced by their [Enterprise API](https://developer.twitter.com/en/products/twitter-api/enterprise). But since I don't have access to that I've never known what it was.

Would you be willing to add a little unit test that is skipped if the GNIP environment isn't present? You can see an example of this at the bottom of the test file for the Premium API tests.

epicfaace:
Sure, will do.

epicfaace:
@edsu I've added that test; however, I've temporarily lost access to GNIP so I can't actually verify that it works.

Some further things beyond this PR that I think would be nice:
- interactively configure GNIP credentials by running `twarc configure --gnip_auth`
- adding mock API responses so that we can run GNIP / premium / other tests on CI without needing to use API keys

Let me know if you are good with these proposals, and I'll probably add them in other PRs when I get some time.

edsu:
Awesome, thanks for doing that @epicfaace -- if you don't mind too much lets wait until you can verify the tests work before merging this. Maybe someone else with access to the full archive can take a look?

epicfaace:
Okay, I was just able to verify that the test works:

image

edsu:
So awesome, thanks for all this work @epicfaace it is super appreciated!

edsu:
Ok thais was just released as part of v1.12.0. Since it didn't add the 30 day Gnip endpoint I might change things in the future. But I think this is very nice to have for the few people who have access to this service.

epicfaace:
Great, thanks!

PR #365, opened 2020-11-27
Stdin hydrate

nikolayay:


edsu:
I've wanted this functionality myself for some time. So thanks for working on it and submitting the PR! Did you notice how many files are changed in this PR though? I'm having trouble understanding what has happened there. Do you understand it?

igorbrigadir:
@nikolayay it would help if you [rebase your PR](https://dev.to/matks/what-it-means-to-rebase-a-pull-request-submitted-on-github-5717), i wanted this too and had a look and as far as i can tell this is the substantive change: https://github.com/DocNow/twarc/pull/365/commits/4dca8e5e7d13e1ba1040e05daef09b7abe4d163f

I would also like this feature but i think there's 2 cases to consider:

`cat tweets.txt | twarc hydrate`

for reading from STDIN

and 

`cat tweets.txt | xargs twarc hydrate`

which is different, maybe it would help to make this feature clearer: https://stackoverflow.com/a/11853329/11090908

Which one is more appropriate depends entirely on how twarc should ideally process things. I don't have a strong opinion on either, but i think it's important to be clear about how you can use it (to an end user)

Currently this PR reads a single string from STDIN and splits it on a comma, which will not work for piping things and is not usually how you'd expect to interact with a command line tool like this.

Hope that helps!

PR #377, opened 2021-02-09
V2 proposed

SamHames:
@betsybookwyrm  and I have put together some of our thoughts based on our last meeting. Comments appreciated!


edsu:
Thanks for dealing with the churn!

edsu:
PS. I merged this into the v2 branch instead of main.

PR #388, opened 2021-02-20
v2: Add flattened tweet format function

igorbrigadir:
This adds a function that takes an entire page of results, and inserts all `includes` inline into tweets.

### DIFF
--- a/twarc/expansions.py
+++ b/twarc/expansions.py
@@ -1,9 +1,11 @@
 """
 This module contains a list of the known Twitter V2+ API expansions and fields for
-each expansion.
+each expansion, and a function for "flattening" a result set,
+including all expansions inline
 
 """
 
+from collections import defaultdict
 
 EXPANSIONS = [
     "author_id",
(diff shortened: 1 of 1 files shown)


edsu:
Thanks @igorbrigadir this is perfect. I'll add some tests for it, and try to integrate it with @SamHames' sample method.

edsu:
I did end up integrating it into the Twarc2.sample. I think I found a bug in the "geo" handling so an extra set of eyes on what I did would be helpful.

igorbrigadir:
Oh thanks! I'll have a look again, I had some tweet examples to add to the repo to use as tests so those should work.

There's also a geo thing with standard and academic access where it's not returned in standard, so there might be a check needed for that, but i need to double check that.

edsu:
I like the idea of crafting the tests to use live data rather than canned where possible.

igorbrigadir:
> I like the idea of crafting the tests to use live data rather than canned where possible.

Sure, in that case - maybe i can see if i can find some queries to "search" so that they have a high probability of hitting some tweets with geo / media expansions and other features like that.

PR #393, opened 2021-02-28
v2 improved expansions

igorbrigadir:
I changed the method that did the expansions to be more generic and changed how polls are expanded (there can only be 1 poll per tweeet, so it didn't make sense to have them as a list).

Despite now being recursive, the new function should be simpler.

Also stopped removing the includes from the request, so now the function is additive only as advertised.

### DIFF
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -138,7 +138,7 @@ def test_sample_flattened():
 
         if "geo" in tweet:
             assert tweet["geo"]["place_id"]
-            assert tweet["geo"]["place_id"] == tweet["geo"]["place"]["id"]
+            assert tweet["geo"]["place_id"] == tweet["geo"]["id"]
             found_geo = True
 
         if "entities" in tweet and "mentions" in tweet["entities"]:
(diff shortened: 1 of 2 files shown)


igorbrigadir:
I noticed the geo test failing - not sure if that's my mistake, but geo is only expanded if that `place` is included in `includes`.

It's possible to have an unexpanded geo object, now that expansions are done recursively on everything, including the `includes` tweets. Gotta double check this though.

igorbrigadir:
I had to change the test after all - the first time around it was appending a `place` inside `geo` but now it's merging it because a tweet can only have 1 place attached to it.

The original, unflattened geo object is like this for example:

```
{
  "place_id": "28edff73b28c1a74"
}
```

Now the "flattened" one is like this:

```
{
  "place_id": "28edff73b28c1a74",
  "name": "South Miami",
  "id": "28edff73b28c1a74",
  "full_name": "South Miami, FL",
  "country_code": "US",
  "place_type": "city",
  "country": "United States",
  "geo": {
    "type": "Feature",
    "bbox": [
      -80.306913,
      25.689337,
      -80.284859,
      25.734075
    ],
    "properties": {}
  }
}
```

edsu:
Nice, thanks for this @igorbrigadir!

PR #408, opened 2021-03-20
v2: fix metadata not being included on first page

igorbrigadir:
For paginated results, `__twarc` metadata was not being included on the first page of the results.

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -443,6 +443,12 @@ class Twarc2:
         page = self.get(*args, **kwargs).json()
 
         url = args[0]
+        
+        if self.metadata:
+            page["__twarc"] = {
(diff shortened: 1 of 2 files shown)


igorbrigadir:
I just noticed too, when using `--flatten` the `__twarc` metadata is also missing. It might be good to add that in too, i think copying it into the object in `expansions.py` is a good place for it?

edsu:
Thanks for noticing this, I wanted to check that it was happening!

edsu:
Also thanks for the example of using `enumerate`. That's something I will use in the future :-)

edsu:
I'm just now noticing @SamHames use of` `enumerate` in the tests, which is something I will use in the future!

PR #409, opened 2021-03-21
move all metadata appending to single function

igorbrigadir:
This is a small change just to make it so that there's one place where the extra metadata additions are defined.


edsu:
I like this since it ensures that the __twarc metadata is consistent.

PR #410, opened 2021-03-22
mkdocs documentation

edsu:
This is an initial pass at generating documentation using mkdocs and publishing at readthedocs.


edsu:
The mkdocs configuration is light enough that it would be easy to replace with something else if needed. So I'm going to go ahead and merge this.

PR #416, opened 2021-03-27
Minor fix for expansions

igorbrigadir:
Fix an edge case if running expansions on a previously expanded result, for example, if running the same command twice, it shouldn't break, it should just do the same thing.

### DIFF
--- a/twarc/expansions.py
+++ b/twarc/expansions.py
@@ -166,11 +166,11 @@ def flatten(response):
                 includes_media[media_key] for media_key in payload["media_keys"]
             )
 
-        if "poll_ids" in payload:
+        if "poll_ids" in payload and len(payload["poll_ids"]) > 0:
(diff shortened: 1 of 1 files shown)


edsu:
It wasn't clear to me why this adjustment was needed for only these includes. Can you say anything more about that?

igorbrigadir:
Yeah it's more "defensive" changes - this error may crop up if for example, you did your own changes / merges to the json using expansions.py in a separate script not via the command line, and then wanted to use the command line anyway, or used https://pypi.org/project/searchtweets-v2/ or something else like that.

It's needed because when flattening, it will place an empty `{}` dict if it's missing an include, which if you run again, will fail because it exists, but doesn't have the expected id or anything. 

This makes the flatten command idempotent, which is always a good thing to have.

igorbrigadir:
Separately, it seems like the failing test `test_hydrate` might need to be updated with new IDs since a bunch of them got deleted just like the comment said!

PR #424, opened 2021-04-07
add twarc version to metadata

igorbrigadir:


edsu:
Thanks gor this! You read my mind, since this was my original impetus for moving the version to its own module and adding the command. But I didn't follow through. I'm glad you did :-)

PR #439, opened 2021-04-17
fix flatten command overwriting input file #436

### LINKED ISSUE #436: twarc flatten errors
Leaving this here to investigate later: when running

`twarc2 flatten file.json file.json`

(Specifying the same file)

Does it overwrite with an error? I think this should throw an error or auto rename the output file or something. We shouldn't have any potentially destructive commands I think. 

This is coming from a DM, so I'll check it out in code later.

igorbrigadir:
Fixes #436 and shows a warning and writes to a different file instead.

Doesn't write the warning out if using pipes, so `cat tweets.json | twarc2 flatten` still works

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -359,8 +359,8 @@ def flatten(infile, outfile):
     "Flatten" tweets, or move expansions inline with tweet objects.
     """
     if (infile.name == outfile.name):
-        outfile.name = outfile.name + '_flattened.jsonl'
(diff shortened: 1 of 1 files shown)


edsu:
Did you consider simply printing an error? What if _flattened already exists?

igorbrigadir:
Ah yes good point - maybe i'll just fail the command instead.

igorbrigadir:
It's still possible to break an input file if you `twarc2 flatten tweets.json > tweets.json` but that's outside of our control.

PR #446, opened 2021-04-29
For full archive search, sleep in between pages to avoid the 1 reques…

SamHames:
…t/second limit.


edsu:
This looks perfect @SamHames ... you covered the important part by sleeping for the search/all endpoint but not for search/recent. I really liked how you incorporated the timing in between the yields that was super cool. And I got to learn about `time.monotonic()`! :clap:

PR #448, opened 2021-04-29
fix #447 missing JSONDecodeError

### LINKED ISSUE #447: missing JSONDecodeError import
JSONDecodeError is not imported in decorators.py, so when an error happens, Requests throws either a `json.decoder.JSONDecodeError` or, `simplejson.errors.JSONDecodeError` which is ambiguous.

https://github.com/DocNow/twarc/blob/main/twarc/decorators.py#L187

This causes a "NameError: name 'JSONDecodeError' is not defined" for a few people, for example #441 which won't solve their problem but gives it a weird message.

igorbrigadir:
Both `json.decoder.JSONDecodeError` and `simplejson.errors.JSONDecodeError` are `ValueError` so this should fix it.

### DIFF
--- a/twarc/decorators.py
+++ b/twarc/decorators.py
@@ -184,7 +184,7 @@ class cli_api_error():
                     msg = result['title']
                 else:
                     msg = 'Unknown error'
-            except JSONDecodeError:
+            except ValueError:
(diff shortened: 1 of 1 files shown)


edsu:
Thanks for having my back @igorbrigadir!

PR #451, opened 2021-04-30
Check for since-id use when assuming start-time

igorbrigadir:
If you try to use `--since-id` and `--until-id` instead of `--start-time` and `--end-time` it will fail with 
`⚡ Invalid use of 'since_id' or 'until_id' in conjunction with 'start_time' or 'end_time'.` because it will set start time to the 2006 date.

For example:

```
twarc2 search --since-id 947277536320405504 --until-id 966797201685938176 --archive "..."
```

not setting `--start-time` if `--since-id` is set fixes this.


edsu:
Nice catch @igorbrigadir!  I just released that to PyPI as 2.0.12.

PR #455, opened 2021-05-11
Allow twarc to run as module

igorbrigadir:
following on from https://github.com/DocNow/twarc/issues/441#issuecomment-832876616 

It's possible to make it so that these commands are equivalent:

```
twarc2 version
python -m twarc2 version
twarc2.exe version
```

```
twarc version
python -m twarc version
twarc.exe version
```

I'm not sure how generally useful this is but i found it helped because i could run a profiler (slightly) easier, for example:

```
python -m cProfile -s cumulative -m twarc2 csv big_test.jsonl big_test.csv
```

It's not a terribly critical improvement either way, but it was a very small change with some boilerplate files.

### DIFF
--- a/twarc/__main__.py
+++ b/twarc/__main__.py
@@ -1,0 +1,5 @@
+import twarc.command
+
+if __name__ == "__main__":
+    twarc.command.main()
+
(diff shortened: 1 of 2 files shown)


edsu:
That is super elegant @igorbrigadir!

PR #467, opened 2021-05-21
Fix misleading log messages in timeline and search

SamHames:
All data is processed when the API indicates there are no more
pages to retrieve, not when we hit an empty page of results. A
page of results can be empty when tweets are deleted.

As an addendum, also fix a subtle problem that might lead to
exceeding the 1 request/s rate limit for search when there are
lots of empty pages.

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -133,17 +133,19 @@ class Twarc2:
                 count += len(response['data'])
                 yield response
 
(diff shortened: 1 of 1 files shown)


edsu:
Thanks for taking a look at this @SamHames -- I think you found why this particular timeline was triggering an immeidate rate limit.

PR #475, opened 2021-06-03
Stop hiding input with getpass

igorbrigadir:
Hiding the token in the terminal on input is *way* more problems than it's worth. Considering we're storing the tokens in plain text anyway, there is no advantage of hiding the input in the terminal - it only serves to confuse people.


edsu:
Totally agree @igorbrigadir thanks!

PR #479, opened 2021-06-13
Don't show bytes b prefix

hugovk:
```sh
python utils/source.py tweets.jsonl > sources.html
```

# Before

image

# After

image

As only Python 3.3+ is supported, no need to support Python 2.

I've only tested on macOS, is extra handling needed for Windows?

Also update homepage URL.

And is it time to only [support 3.6](https://endoflife.date/python)? Could then use some newer things like f-strings.


igorbrigadir:
Thanks for this too!

PR #480, opened 2021-06-13
wall.py: Fix username regex and show likes

hugovk:
https://help.twitter.com/en/managing-your-account/twitter-username-rules says:

> A username can only contain alphanumeric characters (letters A-Z, numbers 0-9) with the exception of underscores, as noted above. Check to make sure your desired username doesn't contain any symbols, dashes, or spaces.

So that's `[A-Za-z0-9_]+`

# Before

The existing one has a couple of problems:

* The first username in a tweet isn't linked, for example b3703742 in the second tweet, similarly in the third and sixth tweets
* Colons are included in link, for example the first links `b3703742:` to `https://twitter.com/b3703742:`, similarly in the fourth and fifth

image

# After

Fix those and show likes:

image

Also update twarc and HTTPS URLs and upgrade for Python 3+.

### DIFF
--- a/utils/wall.py
+++ b/utils/wall.py
@@ -168,7 +168,7 @@ for line in lines:
         start, end = url['indices']
         t['text'] = t['text'][0:start] + a + t['text'][end:]
 
-    t['text'] = re.sub(' @([^ ]+)', r' <a href="https://twitter.com/\g<1>">@\g<1></a>', t['text'])
+    t['text'] = re.sub('@([A-Za-z0-9_]+)', r'<a href="https://twitter.com/\g<1>">@\g<1></a>', t['text'])
     t['text'] = re.sub(' #([^ ]+)', r' <a href="https://twitter.com/search?q=%23\g<1>&src=hash">#\g<1></a>', t['text'])
 
     html = """
--- a/utils/wall.py
+++ b/utils/wall.py
@@ -185,10 +185,7 @@ for line in lines:
     </article>
     """ % t
 
-    if sys.version_info.major == 2:
-        print(html.encode('utf8'))
-    else:
-        print(html)
+    print(html)
 
 print("""
 
--- a/utils/wall.py
+++ b/utils/wall.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 
 """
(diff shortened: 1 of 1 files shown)


igorbrigadir:
This looks good to me! Thanks! 

Since it's a small change that only affects the script and nothing else it should be ok to merge straight in.

I think for brand new things, it's better to use twarc2 plugins https://twarc-project.readthedocs.io/en/latest/plugins/ but the old scripts like this one in `utils` are still useful for people and it's good to maintain them like this.

hugovk:
Yes, I should probably look into twarc2 at some point, but I'm not twarcing so often and already have all the keys set up for twarc(1).

It might be an  idea at some point to switch the docs to twarc2 first, with a link to twarc?

https://twarc-project.readthedocs.io/en/latest/

And some more detailed instructions for configuring twarc2 would be useful.

Thanks!

igorbrigadir:
Yeah - the docs can always be improved! I guess while v1.1 API is still alive twarc will be the first one, but we can figure that out later.

> And some more detailed instructions for configuring twarc2 would be useful.

You mean as a library or as a command line tool?

`twarc2 configure` should work but i was also thinking of adding `--bearer` command line option to pass in auth details on the fly, without relying on a saved file. But that's maybe a point for a different issue / PR.

hugovk:
As a command line tool.

My path checking the docs was:

Visit https://twarc-project.readthedocs.io/. Spot the link at the top going to [twarc2](https://twarc-project.readthedocs.io/en/latest/twarc2/). Then it jumps right into the command reference guide. Compare with the twarc docs which start with installation and quickstart guide.

Checking the twarc2 reference guide, https://twarc-project.readthedocs.io/en/latest/twarc2/#configure tells:

```console
$ twarc2 configure
Please enter your Bearer Token (leave blank to skip to API key configuration):
```

But then I didn't know what a Bearer token is. Looking around, I see there's some info at https://twarc-project.readthedocs.io/en/latest/twitter-developer-access/, I guess this is what I need to follow? Perhaps link to this from the configure command, or made more prominent in some way?

Thanks!

igorbrigadir:
Yes, this is a good point - i'll see if i can add some more details to those docs - thanks again!

PR #486, opened 2021-06-19
Apply python black formatting

igorbrigadir:
There are no changes here - only applying python black formatting to files.

Any opinions on this? It's very handy to have `black` auto formatting code in my opinion. I'm open to other formatting suggestions.


edsu:
Thanks for raising this @igorbrigadir I've noticed that we have slightly different styles in places and it can make things hard to read at times. Can we set things up so that the test suite ensures code is formatted properly?

igorbrigadir:
Yeah that's not a bad idea!

You mean as a check? Or some other way? I added a github action that seems like it would work.

edsu:
So an action that's distinct from the test suite? I think it would be nice to be able to notice the problem with a `python setup.py test` ?

igorbrigadir:
Oh! yes - that's a better idea - i added using https://pypi.org/project/pytest-black/ - i think this is all that's needed?

edsu:
This is a big improvement! :heart_eyes_cat:

PR #487, opened 2021-06-20
Change flatten test to use faster search api method instead

igorbrigadir:
This was really annoying me because it was taking so long to run. I changed the flatten test to use the Recent Search API instead. The query usually gets everything within ~300 tweets.

```
pytest test_twarc2.py -k 'test_flattened'
```

to run just this.

I think this preserves the "live test" while also being efficient with the quota.

### DIFF
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -319,11 +319,11 @@ def test_follows_username():
 
 def test_flattened():
     """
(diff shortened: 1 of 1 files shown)


edsu:
I like this a lot, awesome idea!

Issue #399, opened 2021-03-04
Support Batch Compliance Endpoints


igorbrigadir:
There is a beta for Batch compliance endpoints https://developer.twitter.com/en/docs/twitter-api/compliance/batch-tweet/api-reference

I think i still have access to this so it would be good to implement these for later. 

These endpoints allow you to upload a zipped file of dehydrated tweets, and get back compliance objects (tweet deletions, scrub geo events etc.)

PR #488, opened 2021-06-20
Batch compliance implementation

igorbrigadir:
For #399

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -946,7 +946,7 @@ class Twarc2:
             payload["name"] = job_name
 
         result = self.client.post(
-            "https://api.twitter.com/2/tweets/compliance/jobs", json=payload
+            "https://api.twitter.com/2/compliance/jobs", json=payload
         )
 
         if result.status_code == 200:
@@ -957,7 +957,6 @@ class Twarc2:
             return result
         else:
             raise ValueError(f"Unknown response from twitter: {result}")
-
 
     def _id_exists(self, user):
         """
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -905,8 +905,8 @@ class Twarc2:
     @requires_app_auth
     def compliance_job_get(self, job_id):
         """
-        Returns a compliance job. 
-        
+        Returns a compliance job.
+
         Calls [GET /2/tweets/compliance/jobs/{job_id}](https://developer.twitter.com/en/docs/twitter-api/compliance/batch-tweet/api-reference/get-tweets-compliance-jobs-id)
 
         Args:
@@ -921,9 +921,7 @@ class Twarc2:
         if result.status_code == 200:
             result = result.json()
         else:
-            raise ValueError(
-                f"Error from API, response: {result.status_code}"
-            )
+            raise ValueError(f"Error from API, response: {result.status_code}")
         if "data" in result:
             return result
         elif "error" in result:
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -917,7 +917,13 @@ class Twarc2:
         """
         result = self.client.get(
             "https://api.twitter.com/2/tweets/compliance/jobs/{}".format(job_id)
-        ).json()
+        )
+        if result.status_code == 200:
+            result = result.json()
+        else:
+            raise ValueError(
+                f"Error from API, response: {result.status_code}"
+            )
         if "data" in result:
             return result
         elif "error" in result:
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -775,6 +775,19 @@ class Twarc2:
             else:
                 raise ValueError(f"No such user {user}")
 
+    @requires_app_auth
+    def compliance_job_list(self, start_time, end_time, status):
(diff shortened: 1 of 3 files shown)


edsu:
This is going to be AWESOME for people who have tweet id datasets and don't want to run/manage the hydrate themselves. Do you want this to be merged before the stubbed out create and get subcommands are finished?

igorbrigadir:
I was kinda delaying this because I was trying to get a good progress bar implementation in first, but I might do a more rudimentary thing just to get this working fully

edsu:
Ok, no rush. I agree the progress bar is pretty important for this functionality.

igorbrigadir:
The batch compliance endpoint alpha is on hold while they get ready to make this generally available - no timeline but i'll finish this when the API works again.

edsu:
Is this work ready yet?

igorbrigadir:
Yep - almost - i had to go back and change some things because the API changed (it now supports users and tweets)

igorbrigadir:
I went back and forth a bit with the commands, but in the end i ended up with a slightly simplified approach than the reference app: https://github.com/twitterdev/compliant-client/blob/main/apps/compliant-client.py

There are 4 commands under `twarc2 compliance-job`:

```
Commands:
  create    Create a new compliance job and upload tweet IDs.
  download  Download the compliance job with the specified ID.
  get       Returns status and download information about the job ID.
  list      Returns a list of compliance jobs by job type and status.
```

To create a compliance job, 

```
twarc2 compliance-job create --job-type tweets tweet_ids.txt
```

`--job-type` is mandatory

Optionally you can set a descriptive name, and wait for it to finish:

```
twarc2 compliance-job create --wait --job-type users --job-name "100 user ids" user_ids.txt
```

(user jobs tend to finish faster for some reason, tweet jobs can take up to 20 or 30 mintutes - i haven't tested these times extensively yet with different datasets)

Downloading jobs has an optional `--wait` that will show a progress bar ticking every minute showing the job status until it downloads (same thing for `--wait` when using `create`). File output name is optional, default is `type_compliance_id.json` like `users_compliance_1437534916447715338.json`. 

```
twarc2 compliance-job download --wait 1437534916447715338 foo.json
```

Listing and getting a specific job by ID can have `--verbose` output or `--json-output` and if you don't specify `--job-type` both are output.

eg:

```
twarc2 compliance-job list
```

or

```
twarc2 compliance-job list --status "in_progress"
```

You can't list "expired" jobs with `--status` but they appear in the list of all jobs.

The output looks like:

```
📃 Type: "tweets" ID: "1437529012717375488" Name: "Job" Status: "expired" 
📃 Type: "tweets" ID: "1437533911471493128" Name: "test" Status: "complete" 
📃 Type: "users" ID: "1437496928955215874" Name: "Job" Status: "expired" 
📃 Type: "users" ID: "1437534916447715338" Name: "100 user ids" Status: "complete"
```

To get a single job for example (`--json-output` optional): 

```
twarc2 compliance-job get 1437533911471493128 --json-output
```

Running `twarc2 compliance-job get` without `--json-output` on an already completed job will prompt to download the results.

Not implemented features:

I did not implement resumable uploads yet. We'll see how it fares. I have a feeling they maybe necessary. It's only slightly more awkward to do, you are only allowed 1 job per type (users or tweets) to be active at any one time.

An important difference with other commands, is there is *no* STDIN and STDOUT. It got awkward and was easier to implement assuming already created files.

The input file is not automatically processed - it expects an already made file of ids, but i put in a copy pastable command as a helpful error when trying to create a compliance job with a dataset of tweets:

```
$ twarc2 compliance-job create raw.json --job-type tweets
🙃 The file raw.json does not contain a list of IDs. Use:
 twarc2 dehydrate --id-type tweets raw.json output_ids.txt
to create a file with tweets IDs.

$ twarc2 compliance-job create raw.json --job-type users
🙃 The file raw.json does not contain a list of IDs. Use:
 twarc2 dehydrate --id-type users raw.json output_ids.txt
to create a file with users IDs.
```

I'm open to changing any and all of this around. Maybe making `--wait` the default? and changing it to optionally `--no-wait`?

SamHames:
I've just been through a test run of the workflow, using the --wait option for 1 pass through: it works well. I think this is a good way to navigate the complexity of the batch workflow. 

Haven't had a chance to look through the code but will take a look later to make sure I understand the workflow.

Minor UI notes:

To me I think the --wait by default option is preferred. These kind of jobs are the kind of things that I start and then forget, so the single command workflow from files to results is what I'd be demonstrating to other people.

Since the job type is mandatory, it might better for it to be a positional argument, rather than a flag:
> twarc2 compliance-job create [OPTIONS] JOBTYPE INFILE [OUTFILE]

SamHames (approved):
Looks good to me apart from the suggestions for CLI.

igorbrigadir:
Yeah! the job type as a positional argument and wait by default is better - i'll add that in.

igorbrigadir:
I think this is good to go too.

igorbrigadir:
(I didn't bump the version with this but since there's other stuff being merged that should be fine)

PR #491, opened 2021-06-21
Flatten Refactor

igorbrigadir:
This is just missing some more tests. I changed the ensure_flatten function to hopefully be easier to understand and use. Before it would fail if presented with a list of user objects, because it relied on checking for `author` field - which won't exist in user objects.

### DIFF
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -405,7 +405,7 @@ def test_flattened():
 
 
 def test_ensure_flattened():
-    resp = next(T.search_recent("twitter"))
+    resp = next(T.search_recent("twitter", max_results=20))
 
     # flatten a response
     flat1 = twarc.expansions.ensure_flattened(resp)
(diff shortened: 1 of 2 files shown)


igorbrigadir:
I think this works for all types of data now, but may need a second pair of eyes / some more examples to test.

edsu:
This looks good to me @igorbrigadir although I haven't looked closely at it. If the test_flatten() unit test passes that's a good sign. I wonder if we might need another test case for passing in User objects (I didn't realize it could work with those).

igorbrigadir:
Ah - i did find a problem - looks like it is better to let it pass without error if there's `data` but no `includes`. `flatten()` can handle this just fine - and it's a valid API response.

For example:

```
twarc2 following overflow64 | twarc2 flatten
```

should work.

igorbrigadir:
Ok, that took slightly more changes - so now, i think it finally covers all cases:

A single response where data is a dict (eg: stream)
A single response where data is a list (eg: search)
A list of responses where data is a dict (eg: using readlines() or something)
A list of responses where data is a list 
An already flattened object
A list of already flattened objects

ensure_flatten will always return a list of objects, and always attempt to expand anything that has `data` - regardless of what's inside includes or not, so it will work for both tweets and users, eg:

```
twarc2 following overflow64 | twarc2 flatten
```

PR #492, opened 2021-06-22
improve v2 _ensure_user_id's handling of numeric screen names

cosmicoptima:
_ensure_user_id currently returns any integer or numeric string unchanged, which means that any method that uses it will mishandle screen names that consist entirely of digits.

I changed its behavior in the case of a string that is a valid but nonexistent ID, but that _is_ an existent screen name. Otherwise nothing should have changed.

Some examples/test cases (is this too minor to add unit tests?):

```
In [3]: api._ensure_user_id("parafactual")
Out[3]: '1252277511746183168'

In [4]: api._ensure_user_id(Out[3])
Out[4]: '1252277511746183168'

In [5]: api._ensure_user_id(Out[3][:-1]) # doesn't exist
Out[5]: '125227751174618316'

In [6]: api._ensure_user_id("Twitter")
Out[6]: '783214'

In [7]: api._ensure_user_id("783214") # both screen name and ID
Out[7]: '783214'

In [8]: api._ensure_user_id("42069") # screen name, not ID
Out[8]: '17334495'
```

### DIFF
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -458,6 +458,26 @@ def test_ensure_flattened():
         twarc.expansions.ensure_flattened([[{"data": {"fake": "list_of_lists"}}]])
 
 
+def test_ensure_user_id():
+    """
+    Test _ensure_user_id's ability to discriminate correctly between IDs and
+    screen names.
+    """
+    # presumably IDs don't change
+    assert T._ensure_user_id("jack") == "12"
+
+    # should hold for all users, even if the screen name exists
+    assert T._ensure_user_id("12") == "12"
+
+    # this is a screen name but not an ID
+    # would help to find more "stable" example?
+    assert T._ensure_user_id("42069") == "17334495"
+    # should 42069 passed as int return ID or screen name?
+
+    assert T._ensure_user_id("1033441111677788160") == "1033441111677788160"
+    assert T._ensure_user_id(1033441111677788160) == "1033441111677788160"
+
+
 def test_twarc_metadata():
 
     # With metadata (default)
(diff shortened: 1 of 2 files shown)


igorbrigadir:
Thanks!

This picks up some very interesting edge cases - there's not that many users with all number screen names, but it's great to account for them.

I think the checks for tests won't run automatically here because of the repo settings but they seem to pass for me locally.

cosmicoptima:
they passed (or at least didn't fail) for me as well

igorbrigadir:
Great! Looks good to me! Any new ids `len(user) > 15` should work as before without triggering the extra check, so really it's for edge cases for account created before `2013-01-22` and have all digits as their screen name.

cosmicoptima:
yeah, most of the older accounts I tested had IDs with less than 15 characters; I think there are interesting conflicts between IDs and screen names <1000, but not much I can do about that

edsu:
I'm confused about why testing the length is needed.

cosmicoptima:
Correct me if I'm wrong, but I don't think usernames can be longer than 15 characters, so this would save an API call.

edsu:
Is this not correct then?

https://twitter.com/TwitterSupport/status/928777729180254208

cosmicoptima:
That refers to the display name, not the screen name. (A screen name can't contain emojis.)

edsu:
Ok, thanks! I learned something new today.

cosmicoptima:
Maybe I have the terminology wrong. I'm using "screen name" to refer to what comes after the "@" since twarc seems to do this as well.

PR #494, opened 2021-06-23
Handle quoted user ids in twarc2 timelines command

SamHames:
Also this ensures that ensure_flattened is called only once, which
shouldn't be a problem, but might rule out another issue reported
on slack.

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -529,7 +529,14 @@ def timelines(
             try:
                 users = set([t["author"]["id"] for t in ensure_flattened(json_data)])
             except ValueError:
-                users = set(json_data)
+                # If it's not a structure we understand, but it is a nonempty string,
+                # pass it through as a user_id itself. Downstream functions will do
+                # additional validation.
+                if isinstance(json_data, str) and json_data:
+                    users = set([json_data])
+                else:
+                    # Ignore the line if it's JSON but we can't deal with it
+                    continue
 
         except json.JSONDecodeError:
             # Can't decode JSON, just pass straight through as a raw user_id
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -516,36 +516,50 @@ def timelines(
 
     """
     total_count = 0
+    line_count = 0
     seen = set()
     for line in infile:
+        line_count += 1
         line = line.strip()
         if line == "":
+            logging.warn("skipping blank line on line %s", line_count)
             continue
 
-        users = []
+        users = None
         try:
-            # Handle both quoted user_ids and json tweet data
+            # first try to get user ids from a flattened Twitter response
             json_data = json.loads(line)
             try:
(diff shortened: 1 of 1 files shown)


edsu:
Looks good!

twarc/command2.py
```diff
@@ -518,12 +524,16 @@ def timelines(
...
         users = []
         try:
-            data = ensure_flattened(json.loads(line))
-            users = set([t["author"]["id"] for t in ensure_flattened(data)])
+            # Handle both quoted user_ids and json tweet data
+            json_data = json.loads(line)
+            try:
+                users = set([t["author"]["id"] for t in ensure_flattened(json_data)])
+            except ValueError:
+                users = set(json_data)
```
edsu, line 532: I see this was introduced to guard against usernames that are quoted? Isn't there a danger here that any JSON that is passed in will be sent along? Maybe we could check that `json_data` is a string and log an error ad skip if it's not?

igorbrigadir:
I ran into a similar, related issue - but in my case it was a file of IDs extracted from a CSV, so it had:

```
123456
""
122343
123164
```

Missing entries were `""` quoted and it broke on that.

Maybe something simpler in this case like `int('"10" '.replace('"','').strip())` would work?

I really like, and think we should add the support for passing a file of results to this command, and let it extract the ids, but i think it's a bit ambiguous right now - because you can have a dataset of tweets, or a dataset of users, so for a dataset of users `t["author"]["id"]` won't work, it would have to be `t["id"]`. But this can be detected - users will always have `username` when tweets wont.

It's a bit more extra checks and work but it might be worth it. Checking for things like these (after json.loads) is very fast, so it's not going to matter. (check if `'username' in json_data`, extract `t['id']` otherwise `t["author"]["id"]` otherwise just the number is what i mean)

Hope that makes sense! But yeah - this is a good idea.

twarc/command2.py
SamHames: That's a good point, I've fixed that now.

SamHames:
I think this is enough for now?

I do agree with Igor's point about the ambiguity, but maybe the solution to that is to decouple the user_id/tweet_id extraction functionality from this particular function,  as there are other places that's useful. For example:

> twarc2 extract_user_ids infile.jsonl | twarc2 timelines - timelines.jsonl

igorbrigadir:
How about

```
twarc2 ids --users infile.jsonl | twarc2 timelines - timelines.jsonl
```

instead?

But yeah - the rest looks good to me!

edsu:
Hmm yeah, I wonder if twarc ids should be brought into twarc itself. Then --users could be added. I do kind of like not having the extra step of extracting the user ids, but maybe that's lazy thinking.

I added a bit of logging and fixed a small bug where a set was created out of a string and will merge in. 

I am left feeling that these subcommands are getting complicated enough to deserve tests of their own.

PR #498, opened 2021-06-24
add counts unit test

igorbrigadir:
Forgot the test - this passes locally but the repo still has issues with the stream test.


edsu:
Thanks! I don't know what's up with the repo stream test. It works fine locally.

PR #499, opened 2021-06-25
Proposed backoff behaviour for 503 errors on the academic archive search

SamHames:
This is WIP, and I probably won't have time to test or make any changes in the next week :( 
It's definitely important enough for a second set of eyes to make sure I'm not doing something dumb before merging.


SamHames:
With the recent changes for the context annotations, I'm thinking this might not be necessary anymore?

igorbrigadir:
Yeah - i'd be inclined to still keep this code though - seems like it would be useful for future 503s that twitter will undoubtedly throw.

PR #502, opened 2021-06-30
Docs tutorials links

### LINKED ISSUE #306: Tutorial?
@alblaine has created a useful [tutorial](https://github.com/alblaine/twarc-tutorial) for using twarc. It might be good to think about how something like this could live in the twarc repository itself, or if the existing docs could be improved to address ease of use.

### LINKED ISSUE #411: Document Common V2 usecase: Crawl archive tweets, flatten, export to CSV
Add instructions and update https://github.com/DocNow/twarc/blob/main/utils/json2csv.py or add a new one that does a commonly required operation: Crawl some tweets, process the expansions, and export a CSV for further data analysis in another tool.

igorbrigadir:
fix #306, fix #411 

This adds a docs page to other tutorials. 

Are there any more i missed?

### DIFF
--- a/docs/tutorials.md
+++ b/docs/tutorials.md
@@ -1,0 +1,33 @@
+# Twarc Tutorials
+
+Documentation here is largely auto generated from the code, which may not always be the most user friendly. Others have written great totorials on using twarc, and these are linked here.
+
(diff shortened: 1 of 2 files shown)


igorbrigadir:
Going to add more as I find them

PR #513, opened 2021-07-30
Add progress bar to counts command

igorbrigadir:
I think it's useful to have that in `twarc2 counts` when getting counts for a long period with `--archive`. This works the same as the search progress bar (hidden if using piped output, based on time spans and shows a total tweets count while running)


edsu:
Awesomeness!

PR #520, opened 2021-08-17
fix #519 progress bar start time

### LINKED ISSUE #519: Incorrect time range in progress bar
After #518 the desired progress bar behavior is no longer working for `search`. For example:

    twarc2 search --archive "from:DarpaDan" out.jsonl

should display a search bar going back to the beginning of Twitter `/15 years` instead of /6 days`.

While fixing this it is probably worth spot checking [these manual searches](https://github.com/DocNow/twarc/pull/490#issuecomment-873450447) to ensure they are all behaving as intended.

igorbrigadir:
This puts the `start_time` back into the command line, so now the new behavior is:

In the command Line if `--archive` is set, and no `since_id` or `start_time` is specified, it will assume you want everything since 2006. The Library is now the same as command line, but different to the Twitter API.

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -234,6 +234,12 @@ def _search(
         # default number of tweets per response 500 when not set otherwise
         if max_results == 0:
             max_results = 100  # temp fix for #504
+
+        # start time defaults to the beginning of Twitter to override the
+        # default of the last month. Only do this if start_time is not already
+        # specified and since_id and until_id aren't being used
+        if start_time is None and since_id is None and until_id is None:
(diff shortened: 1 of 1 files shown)


edsu:
Did you want to commit any changes to twarc/client2.py?

igorbrigadir:
Yeah, I just wanted to fix the progress bar for now - I mainly use the command line and while I do think it's good to stick to the API definition, I'd like to see more examples and hear from others that use the twarc client API - now that stuff isn't obviously broken it's more of a preference than a bug

PR #536, opened 2021-09-13
Warn when sleeping too long that the system clock might be inaccurate

SamHames:
Just taking your fix one step further and having an explicit warning message about the system time - it's probably better to fix that rather than always wait 15 minutes :)

Although, maybe we should check for negative wait times as well?


igorbrigadir:
Yes! This is a great idea! I think checking for negative wait times would also make sense

igorbrigadir (approved):

PR #540, opened 2021-09-22
Add a searches function, that executes multiple searches from an inpu…

### LINKED ISSUE #528: CLI entrypoint for running multiple searches
There are a few common use cases that can be boiled down to "run many searches" - think of retrieving the mentions for lots of different accounts, or searching for lots of keywords or phrases that wouldn't otherwise fit into a single search.

It would be nice if there were a `twarc2 searches` command that works similarly to the timelines command, taking an input file with one line per query, and working through the list to retrieve everything.

It would be even nicer if it batched and concatenated queries up to the character limit, to minimise the number of API calls made.

SamHames:
…t file

Closes #528

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -1077,6 +1077,15 @@ def _timeline_tweets(
     "outfile will be a CSV containing the counts for all of the queries in the input file.",
 )
 @click.option(
+    "--combine-queries",
+    is_flag=True,
+    default=False,
+    help="""Merge consecutive queries into a single OR query.
+    For example, if the three rows in your file are: banana, apple, pear
+    then a single query ((banana) OR (apple) OR (pear)) will be issued.
+    """,
+)
+@click.option(
     "--granularity",
     default="day",
     type=click.Choice(["day", "hour", "minute"], case_sensitive=False),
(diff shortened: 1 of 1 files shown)


SamHames:
I'm still not completely sure about the interface to this - but it does let you specify a bunch of searches in an input file, and includes the option to do a dryrun with the counts function first. Some more testing is needed.

SamHames:
Would appreciate a second set of eyes on this one now. I think it's ready to go, but the design adds an inconsistency in command names: we currently have `search` and `counts`, and this add `searches` which does both of those with a command line switch (`--counts-only`). I went with the flag instead of a separate command because I couldn't think of a name that made sense for a counts-driven-by-file command (search -> searches, count->countses?).

The --combine-queries option is meant to allow merging a set of queries using a disjunction, so you make fewer API calls.

edsu:
This looks very thorough @SamHames and works well ... it almost makes me wonder if search should have --counts-only instead of there being a separate command. The `--combine_queries` idea is pretty ingenious! :-)

edsu (approved):
This looks good Sam! I'm curious, did the idea for this come from a user request?

SamHames:
Yeah, the idea originally came from discussion in the Docnow slack - but I kept talking to people where it would save more complicated scripting so thought I'd get it done :)

PR #543, opened 2021-09-22
Line count based progress bar

### LINKED ISSUE #541: Would be handy to have a progressbar that uses the number of lines in the file, not the bytes
FileSizeProgressBar isn't quite right for line based inputs like hydration, timelines and the searches functionality proposed in #540 

Ideally we'd have a LinesInFileProgressBar for those cases.

igorbrigadir:
fix #541

### DIFF
--- a/twarc/decorators2.py
+++ b/twarc/decorators2.py
@@ -197,6 +197,71 @@ class InvalidAuthType(Exception):
     """
 
 
+class FileLineProgressBar(tqdm):
+    """
+    A progress bar based on input file line count. Counts an input file by lines.
+    This tries to read the entire file and count newlines in a robust way.
+    """
+
+    def __init__(self, infile, outfile, **kwargs):
+        disable = False if "disable" not in kwargs else kwargs["disable"]
+        if infile is not None and (infile.name == "<stdin>"):
+            disable = True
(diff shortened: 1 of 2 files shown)


twarc/decorators2.py
```diff
@@ -197,6 +197,82 @@ class InvalidAuthType(Exception):
...
+        def blocks(files, size=65536):
+            while True:
+                b = files.read(size)
+                if not b:
+                    break
+                yield b
+
+        total_lines = 0
+        with open(infile.name, "r", encoding="utf-8", errors="ignore") as f:
+            total_lines = sum(bl.count("\n") for bl in blocks(f))
```
SamHames, line 238: Suggestion 229 - 238 can be replaced with the following, as Python file IO in this case is on demand:

```python
with open(infile.name, "r", encoding="utf-8", errors="ignore") as f:
            total_lines = sum(1 for _ in f)
```

SamHames (commented):
Looks good, thanks!

igorbrigadir:
Yeah! this way is better!

edsu:
Excited to see this!

igorbrigadir:
I tried it on a few examples and it seemed to work for me - gonna merge anyway, and if there's problems i'll fix it before the next release!

Issue #453, opened 2021-05-06
Random sample option


igorbrigadir:
A `--sample N` option for `twarc2 search` to modify retrieval and get around the API limitation of returning all tweets in reverse chronological order.

eg:

```
twarc2 search --archive --sample 10 "example query" output.json
```

To return a 10% sample of results.

Tweet IDs are snowflake ids, which encode a millisecond timestamp. Sampling on the millisecond field is how twitter "samples" 1% streams, by taking a fixed window of millisecond values, which we can emulate here too. Specifying dates as tweet ids using `since_id` and `until_id`, it should be possible to get the millisecond time ranges required to effectively create random sample of tweets for a query.

This won't require any new dependencies, and will purely be based on rewriting the original query parameters to use `since_id` and `until_id` and issuing lots of calls (while staying within the rate limit) so it's maybe not suitable for a plugin, but i'm open to suggestions for deciding to include this in twarc2 vs plugin.

For reference: 

https://github.com/client9/snowflake2time

https://twitter.com/JurgenPfeffer/status/1191172359216078849

https://twittercommunity.com/t/generating-a-random-set-of-tweet-ids/150255

edsu:
I like it! Maybe `--sample .10` or is that too dense?

igorbrigadir:
Yeah - i'd like it to accept a "named" sample strategy or an integer to simplify things.

Some important points on this - sampling by splitting up a query into multiple ones will work technically but practically there's a huge speed trade-off:

The rate limit of 1 request per second, means that sampling in v2 will take 10 requests for every second of data. Which may be fine for filling in gaps but may not be viable for gathering a dataset that spans a long time. It should be fine for sampling a short time range.

SamHames:
Maybe a dumb question: what is the point of recreating Twitter's sampling logic - what are the actual intended use cases? 

I can see lots of use cases for sampling, but I struggle to see how they are satisfied by a slow recreation of the Twitter approach. 

What would be useful to me is something like retrieving one small page of results starting from a randomly selected second within the hour. Each page gives you a point estimate for the tweet rate for that hour, and you could move through a wide time window quickly (approx. 1 week/15 minutes). This would let me narrow down time regions of interest and estimate tweet volumes before collecting everything.

igorbrigadir:
Yeah this is the thing - given how slow it will be, it's only real use is to reconstruct a sample stream if you had gaps due to outages - which is pretty limited i admit.

[...]

PR #566, opened 2021-10-28
Random Sample for Search

igorbrigadir:
For #453, second attempt at #459

Twitter "sample" stream is based on selecting tweets with ids where the millisecond timestamp matches a defined range.

Use `since_id` and `until_id` parameters and snowflake id tricks to simulate a `sample:` operator that samples tweets based on millisecond time windows.

`--sample` command line option can apply to any endpoint that has a since / until id option.

The idea is to is to accept an integer between 1 and 100 to get a sample of n% of tweets, or `--sample gardenhose` or `--sample spritzer` or `--sample v1` (alias for `--sample 1` and `--sample spritzer`) or `--sample v2` which is also a 1% sample but with different sampling windows as far as i can tell.

I still have to make sure my assumptions are correct - but so far the millisecond ranges are like this:

```
        _1% v1 "Spritzer" Sample   [657-666]
        10% v1 "Gardenhose" Sample [657-756]
        10% v1 "Enterprise" Sample [*0*]
        _1% v2 Sample           [*0*]
        _N% v2 Sample           [?]
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
         ACCESS_TOKEN: ${{ secrets.access_token }}
         ACCESS_TOKEN_SECRET: ${{ secrets.access_token_secret }}
         BEARER_TOKEN: ${{ secrets.bearer_token }}
+        SKIP_ACADEMIC_PRODUCT_TRACK: true
       run: python setup.py test
 
     - name: Ensure packages can be built
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # twarc
 
-[![DOI](https://zenodo.org/badge/7605723.svg)](https://zenodo.org/badge/latestdoi/7605723) [![Build Status](https://github.com/docnow/twarc/workflows/tests/badge.svg)](https://github.com/DocNow/twarc/actions/workflows/main.yml)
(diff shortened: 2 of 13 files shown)


igorbrigadir:
Yes, maybe - my current prototype implementation for this is still forthcoming unfortunately.

troyneilson:
> 

Hi digi686. Would you happen to have the code available for your "randomising" loop? I too have academic access and I'm looking to take a sample based on a hashtag search over a 10-year period for sentiment analysis.

edsu:
@igorbrigadir I wonder if it makes sense to release your prototype as a plugin while it is in development?

msa-digi:
> > 
> 
> Hi @digi686. Would you happen to have the code available for your "randomising" loop? I too have academic access and I'm looking to take a sample based on a hashtag search over a 10-year period for sentiment analysis.

Hi @troyneilson, sure! Here's my loop. I put it inside the main function, before defining my query. Since my last comment, I opted for chunks of 2 seconds to reduce the volume of tweets. Hope it helps.

`    for day in tqdm(range(13)):
        for hour in tqdm(range(24)):
            for minute in tqdm(range(60)):
                # Specify the start time in UTC for the time period you want Tweets from
                start_time = datetime.datetime(2018, 11, 1+day, hour, minute, 0, 0, datetime.timezone.utc)

                # Specify the end time in UTC for the time period you want Tweets from
                end_time = datetime.datetime(2018, 11, 1+day, hour, minute, 2, 0, datetime.timezone.utc)
`

troyneilson:
Thanks heaps for that, really appreciated.

PR #569, opened 2021-11-05
Add example configuration and script to build a windows installer

SamHames:
This is a draft of what a windows installer looks like with pynsist.

I think in principle this could be automated to run on releases, with the built installer attached as an artifact to the release itself.

Things I haven't worked out yet:

- code signing, to avoid warnings about untrusted developers
- pynsist requires you specify an entrypoint, which creates a task bar entry, which is definitely odd for a command line program. Maybe that should open a browser and link to the twarc help about running as a command line application?

### DIFF
--- a/installer.cfg
+++ b/installer.cfg
@@ -1,0 +1,26 @@
+[Application]
+name=twarc
+version=2.8.0
+entry_point=twarc.command2:twarc2
+
+[Python]
+version=3.9.7
+bitness=64
+format=bundled
+
+[Include]
+# Requires downloading and/or creating necessary windows wheels locally
+# See windows_installer.sh
+local_wheels = wheels/*.whl
+
+packages=twarc
+
+# Other files and folders that should be installed
+files = LICENSE
+    README.md
+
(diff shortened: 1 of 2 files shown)


igorbrigadir:
This would be great to have! 

Maybe the entry point should launch a command prompt? With `twarc2 --help` or something?

edsu:
Have you compared pynsist with [PyInstaller](https://pypi.org/project/pyinstaller/). I think the latter would build an executable for other platforms, which might be useful?

SamHames:
I did have a look - the main thing pynsist does is create a proper
installer which sets the path and other things needed. Despite the name
pyinstaller only creates an executable that can be packaged by something
else into an installer.

I did miss it the first time around but c-freeze supports creating both
windows msi installers and Mac dmg formats. So I'll take a look at that
next time I get the chance.

I am wary of trying to do too much at once too, especially for platforms
that aren't what I'm usually working on. There's also going to be work
figuring out codesigning and other things 🙈

PR #576, opened 2021-12-03
Make sure rate limits are avoided in loops calling search endpoints

### LINKED ISSUE #575: Rate limit hits when using searches / conversations / file based input
When a loop like https://github.com/DocNow/twarc/blob/51b31cd3fac7e2349ae6b196f28241ad80bc2932/twarc/command2.py#L1448 runs, I think it's possible there's a <1 second between calls when it loops over and processes users / searches that causes a rate limit hit and a 15 min wait, maybe to be on the safe side an extra wait is needed here

Example log where this is happening: https://twittercommunity.com/t/inconsistent-rate-limit-academic-research-full-archive-search/162928/14?u=igorbrigadir

igorbrigadir:
fix #575

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -1662,9 +1662,10 @@ def conversations(
                         log.info(f"reached conversation limit {conversation_limit}")
                         break
 
-                # make sure there's at least a second between calls when calling search in a loop.
-                time.sleep(max(0, 1.05 - (time.monotonic() - last_call)))
-                last_call = time.monotonic()
+                # make sure there's at least a second between calls when calling full archive in a loop.
+                if archive:
+                    time.sleep(max(0, 1.05 - (time.monotonic() - last_call)))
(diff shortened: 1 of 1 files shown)


igorbrigadir:
I think these are the only places where this can happen - anywhere where we read a file and loop over many things making a separate call for each.

igorbrigadir:
I was also thinking maybe this is better fixed in the library? or the decorators where the rate limit sleep happens? (checking to see if there are remaining calls before deciding to sleep for 15 minutes or 1 second - maybe that one should be also added too)

SamHames:
I think this is better fixed in the library, because we don't people writing there own loops to run into the same thing - it's a quick fix, I'll put up a PR in a moment.

igorbrigadir:
Yeah if you have a fix that does this in the library that would be even better!

igorbrigadir:
Closing this in favor of #578

PR #578, opened 2021-12-04
Ensure that every API response item has had the appropriate amount of…

SamHames:
… sleep, even when a generator isn't fully consumed.

This ensures that library and command line users won't hit the 'search/all' rate limit, regardless of whether the entire generator is consumed or not. The bug in #575 is because of the --timeline-limit option - the command line client creates a generator, but doesn't consume all of the pages, which means that we also don't do the necessary sleeping.

This change is simple: make sure that we do the sleeping before we yield a page of results! This potentially slows down some small data collections by a few seconds, because the time between calls is sleep time + processing time of results, but that should be negligible for anything that approaches the 300 calls/15 minute other rate limit.

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -230,6 +230,18 @@ class Twarc2:
         made_call = time.monotonic()
 
         for response in self.get_paginated(url, params=params):
+
+            # Calculate the amount of time to sleep, accounting for any
+            # processing time used by the rest of the application.
+            # This is to satisfy the 1 request / 1 second rate limit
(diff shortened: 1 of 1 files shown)


igorbrigadir (approved):

igorbrigadir:
Ah this is a much better way to do it!

edsu:
Nice work noticing the problem and fixing it. This was a subtle one!

edsu (commented):
Looks good!

PR #584, opened 2022-01-05
Catch 1r/s rate limit and sleep less, instrument overlong sleeps

SamHames:


igorbrigadir (approved):

igorbrigadir:
Tried this and it seems to work! I tried running a long search, and then ran a second instance to try and interrupt it, and it seems to recover handling the 10 second wait and 15 minute wait.

PR #588, opened 2022-01-25
Support likes and retweets endpoints.

### LINKED ISSUE #403: Retweets
I think it would be useful to have a command like this:

    twarc2 retweets {tweet-url}

Which would fetch the retweets by looking up the conversation-id and then issuing the appropriate search. An optional `--archive` parameter could indicate whether to use the Academic Search endpoint.

### LINKED ISSUE #466: Likes endpoint
New likes endpoint gets you 100 latest users who liked a tweet: 
https://twittercommunity.com/t/blocks-and-likes-endpoints-are-now-complete-on-the-twitter-api-v2/154434

I think it's a useful one to have, even if it's limited. 

For example: 
```
twarc2 likes [tweet id] users.jsonl
```

SamHames:
This will add support for the newly useful like and retweets endpoints, that now allow retrieving all of the likes and retweets of a specific tweet, or retrieve a specific users likes. 

This implementation uses the Twitter API naming of these functions (ie what's in the URLs) for consistency.

- [x]  Add client methods to support the new endpoints
- [x]  Add basic CLI layer support
- [x]  Do some targeted testing of the CLI layer/some investigation of the results.
- [x]  Documentation - this will especially need to capture that the liking_users and retweeted_by return user profiles. In the latter case this is also a shame because it doesn't give the *time* information of the retweet.

closes #403 (kind of?)
closes #466

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -876,6 +876,155 @@ def following(T, user, outfile, limit, max_results, hide_progress):
                 break
 
 
+@twarc2.command("liking_users")
+@click.option(
+    "--limit",
+    default=0,
+    help="Maximum number of liking users to retrieve. Increments of 100 or --max-results if set.",
+    type=int,
+)
+@click.option(
+    "--max-results",
+    default=100,
+    help="Maximum number of users (likes) per page. Default is and maximum is 100.",
+    type=int,
+)
+@command_line_progressbar_option
+@click.argument("tweet_id", type=str)
+@click.argument("outfile", type=click.File("w"), default="-")
+@click.pass_obj
+@cli_api_error
+def liking_users(T, tweet_id, outfile, limit, max_results, hide_progress):
+    """
+    Get the users that liked a specific tweet.
+
+    Note that the progress bar is approximate.
+
+    """
+    count = 0
+    lookup_total = 0
+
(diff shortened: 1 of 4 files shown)


SamHames:
Also, is it just me, or does nothing in the API show how many tweets a user has liked? It's visible in the web view of a profile, but apparently it's not the public_metrics field?

SamHames:
Okay, I've done some more testing (and had someone else do some experimentation) and added some more explanatory docs, should be ready to merge if everyone else is happy?

edsu:
This looks *awesome* @SamHames. One thing I've noticed is that we've started to have some inconsistency between snake_case and kebab-case in the cli. I would prefer if we use kebab -- any objections?

twarc/command2.py
```diff
@@ -876,6 +876,155 @@ def following(T, user, outfile, limit, max_results, hide_progress):
                 break
 
 
+@twarc2.command("liking-users")
```
edsu, line 879: I adjusted this to use kebab-case.

edsu (approved):
This looks good! I just adjusted the cli and docs to use kebab-case.

igorbrigadir:
This is great! Thanks again!

PR #592, opened 2022-01-30
Add user-agent headers

### LINKED ISSUE #591: Add a user-agent for twarc
Would be nice if there was a custom user-agent for Twarc (one for v1.1 calls and another for v2 calls)

igorbrigadir:
fix #591 

The user agent is now:

```
twarc/2.9.1 (Linux x86_64) CPython/3.7.5
```

And for v2:

```
twarc/2.9.1 (Linux x86_64) CPython/3.7.5
```

where it's `twarc/2.9.1` is the version from `version.py`, `(OS Architecture)` is there to make sure OS gets parsed, but i didn't want to include a longer more detailed OS version, it should be enough to know if it's Linux, Mac, Windows, including the architecture makes sure it parses in a few user agent parsing things i've tried. `CPython/3.7.5` is the Python interpreter and version which should also be useful.

I think this is ok? not too identifying but not too sparse either. I'm open to suggestions on format / adding / removing things.

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -631,6 +631,7 @@ def test_hydrate():
 @patch("twarc.client.OAuth1Session", autospec=True)
 def test_connection_error_get(oauth1session_class):
     mock_oauth1session = MagicMock(spec=OAuth1Session)
+    mock_oauth1session.headers = {}
     oauth1session_class.return_value = mock_oauth1session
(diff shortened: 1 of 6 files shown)


igorbrigadir:
ICYMI, black got updated and the format changed slightly for some things, so if the next PR keeps failing black tests, black needs to be updated to the latest stable version (it just recently got out of beta)

PR #595, opened 2022-02-07
Avoid exceptions raised by CLI logging for stream/sample commands

SamHames:
For OperationalDisconnect errors, a response is streamed from the
API which causes the logging to fail because it assumes every
entry contains a "data" key.

This just makes sure that the data key is present before trying
to log an archived tweet ID.


igorbrigadir:
Thanks! I just came across this too!

PR #599, opened 2022-02-22
Add the singular user command, analogous to the single tweet command

### LINKED ISSUE #598: twarc2 lookup-user command to retrieve a single user profile
Sometimes, you just need to do a quick lookup to go from a single username to a single user_id. It would be handy in these cases to be able to do, rather than squeeze this specific usecase into the `users` functionality:

`twarc2 lookup-user {username/user_id}` -> single user_profile

or possible even
`twarc2 lookup-user-id username` -> user_id

SamHames:
This adds the user command, which takes a single user id. This is a direct analogue of the single tweet command, so we know have tweet/tweets and user/users.

Closes #598

### DIFF
--- a/twarc/command2.py
+++ b/twarc/command2.py
@@ -1177,23 +1177,22 @@ def users(T, infile, outfile, usernames, hide_progress, **kwargs):
 @twarc2.command("user")
 @command_line_expansions_shortcuts
 @command_line_expansions_options
+@click.argument("name-or-id", type=click.Choice(["name", "id"]))
 @click.argument("user", type=str)
 @click.argument("outfile", type=click.File("w"), default="-")
-@click.option("--username", is_flag=True, default=False)
 @click.pass_obj
 @cli_api_error
-def user(T, user, outfile, username, **kwargs):
+def user(T, name_or_id, user, outfile, **kwargs):
     """
-    Get data for a single user id or username.
+    Get the profile data for a single user by either username or ID.
 
-    By default, user expects a single numeric Twitter user_id, so the
-    following will return the user profile for Twitter's founder:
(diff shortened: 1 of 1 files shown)


edsu:
It looks like id works but username does not?

```
$ twarc2 user jack
⚡ The `ids` query parameter value [jack] is not valid
```

SamHames:
The design mirrors the users command, so there's a  --username flag to
handle that case explicitly.

Not sure that's necessarily the best approach for the long term, but it's
at least consistent with its sibling command.

Given how common that specific use case is do you think it's worth
considering another approach?

edsu (changes requested):
I noticed that looking up by username doesn't work?

```
$ twarc2 user jack
⚡ The `ids` query parameter value [jack] is not valid
```

Maybe that should be fixed or it shouldn't advertise that it can lookup by username?

igorbrigadir:
yes, would be nice to have `twarc2 user jack` work - - very useful for arbitrarily pulling out a user ID on the go!

SamHames:
I added the name/id disambiguation as part of the command to make it clear, as well as updating the docstrings to better align with that.

    Get the profile data for a single user by either username or ID.

    To look up a user by ID:

        twarc2 user id 12

    To look up a user by username:

        twarc2 user name jack

igorbrigadir (approved):

igorbrigadir:
I really like this `"name-or-id"` approach, and the command examples in the docs!

edsu (approved):

edsu:
I like it too! It could be useful for some other commands like`timeline`.

igorbrigadir:
I think this is good to go!

PR #604, opened 2022-03-04
Workaround a limitation in the Twitter API returning empty page

### LINKED ISSUE #602: `twarc2 counts --archive` returning incomplete results
I'm running `twarc2 counts` with the Academic Track of the Twitter API, and for some reason it's returning incomplete results (only going back a few years) with the following query:

```
!twarc2 counts "(\"this is how\" \"not with a\" but) OR (\"not with a bang\") OR (\"not with a\" \"but a whimper\") OR (\"not with a\" \"but with a whimper\") OR (\"this is the way\" \"not with a\" but)" --archive --granularity day --csv > tweets.csv
```
(For reference I'm doing a project about T.S. Eliot's "not with a bang but a whimper.")

When I check twarc.log, I can see that an empty page is returned, which is interpreted as the end of the search:
```
2022-02-28 15:59:21,700 INFO Retrieved an empty page of results
2022-02-28 15:59:21,700 INFO No more results for search.
```

SamHames:
In some cases, the Twitter API will return an empty page of results
from the counts API, without a next_token for pagination and only
a meta value indicating zero tweets in that count. This happens even 
when more results should definitely happen.

This adds a workaround for this, by detecting when the query has
terminated early based on the start-time provided, and restarting
the counts call with the last seen start-time as the new end-time.
This logic won't apply when no start-time is provided (only happens
when using the Twarc client, the CLI will always provide that value).

Hopefully we will get confirmation from Twitter that this is a bug
and will be able to remove this workaround.

Closes #602

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
         ACCESS_TOKEN: ${{ secrets.access_token }}
         ACCESS_TOKEN_SECRET: ${{ secrets.access_token_secret }}
         BEARER_TOKEN: ${{ secrets.bearer_token }}
+        SKIP_ACADEMIC_PRODUCT_TRACK: true
       run: python setup.py test
 
     - name: Ensure packages can be built
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -127,6 +127,10 @@ def test_counts_recent():
     assert 7 <= found_counts <= 8
 
 
+@pytest.mark.skipif(
+    os.environ.get("SKIP_ACADEMIC_PRODUCT_TRACK") != None,
+    reason="No Academic Research Product Track access",
+)
 def test_counts_empty_page():
 
     found_counts = 0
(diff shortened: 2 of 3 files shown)


SamHames:
Whoops - didn't realise the test environment doesn't have full archive access - I don't think there's anyway to test this issue against the recent counts endpoint.

 I will have a look on Monday on how to mark that in pytest for skipping.

edsu:
(twarc) ➜  twarc git:(workaround_counts_API_zero_count) ✗ twarc2 counts "(\"this is how\" \"not with a\" but) OR (\"not with a bang\") OR (\"not with a\" \"but a whimper\") OR (\"not with a\" \"but with a whimper\") OR (\"this is the way\" \"not with a\" but)" --archive --granularity day --csv tweets.csv
100%|█████████████████| Processed 15 years/15 years [16:33<00:00, 354018 tweets total ]
(twarc) ➜  twarc git:(workaround_counts_API_zero_count) ✗

edsu:
Tests are passing now that `SKIP_ACADEMIC_PRODUCT_TRACK` is set for the GitHub action.

Issue #572, opened 2021-11-16
Support new Lists endpoints


igorbrigadir:
The v2 API is officially launched now and has endpoints for reading Lists, which are important to have: 

https://twittercommunity.com/t/announcing-list-lookup-endpoints-for-the-twitter-api-v2/161965

https://developer.twitter.com/en/docs/twitter-api/lists/list-lookup/introduction

igorbrigadir:
Some preliminary ideas on `twarc2 lists` command:

A user on twitter can either create and curate a list, or follow some other public list. I when twarc gets lists of a user, it should get both the "owned" lists and "subscribed" lists as this is more expected, given the lists page of a user (it's a mix of created and followed) eg https://twitter.com/recsyslabs/lists

```
twarc2 lists list [user_id] 
```

should call both: https://developer.twitter.com/en/docs/twitter-api/lists/list-lookup/api-reference/get-users-id-owned_lists and https://developer.twitter.com/en/docs/twitter-api/lists/list-follows/api-reference/get-users-id-followed_lists 

We can also have 

```
twarc2 lists owned [user_id] 
```

```
twarc2 lists followed [user_id] 
```

and 

```
twarc2 lists pinned [user_id]
```

https://developer.twitter.com/en/docs/twitter-api/lists/pinned-lists/api-reference/get-users-id-pinned_lists

separately too. May be better than having a command line switch in `twarc2 lists list` ? 

```
twarc2 lists lookup [list_id]
```

to call list lookup on an individual list https://developer.twitter.com/en/docs/twitter-api/lists/list-lookup/api-reference/get-lists-id

To get followers of a list,

```
twarc2 lists followers [list_id]
```

to call https://developer.twitter.com/en/docs/twitter-api/lists/list-follows/api-reference/get-lists-id-followers

```
twarc2 lists members [list_id]
```

for the members https://developer.twitter.com/en/docs/twitter-api/lists/list-members/api-reference/get-lists-id-members

and 

```
twarc2 lists memberships [list_id]
```

for https://developer.twitter.com/en/docs/twitter-api/lists/list-members/api-reference/get-users-id-list_memberships

and finally 

```
twarc2 lists tweets [list_id] 
```

that calls https://developer.twitter.com/en/docs/twitter-api/lists/list-tweets/api-reference/get-lists-id-tweets

I think grouping command into `twarc2 lists` is better than making them `twarc2 list-members` and `twarc2 lists-tweets` for example, but i'm not strongly attached to this either.

[...]

PR #605, opened 2022-03-08
Added functionality to get membership list

mariosanz92:
This functionality allows to get all the membership list from an specific user ID

-Documentation of endpoint added from twitter API:

https://developer.twitter.com/en/docs/twitter-api/lists/list-members/api-reference/get-users-id-list_memberships

- Endpoint:  https://api.twitter.com/2/users/:id/list_memberships

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -294,6 +294,36 @@ class Twarc2:
 
         log.info(f"No more results for search {query}.")
 
+    def list_memberships(
+        self,
+        id,
+        expansions=None,
+        list_fields=None,
+        max_results=None,
+        pagination_token=None,
+        user=None
+        ):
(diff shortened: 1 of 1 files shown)


igorbrigadir:
Thanks for your contribution! I'm actually in the middle of implementing this for the command line too, so i'm going to go ahead and merge this here, and then add the changes to the `command2.py` and the other list commands according to #572 in a separate PR. This PR takes care of a chunk of it already, thanks!

igorbrigadir:
I added in some changes in `expansions.py` in https://github.com/DocNow/twarc/pull/606/commits/093eb5c72c8ebccdd42bbcce3bf556da0fca6e37 to align it with how the other endpoints are handles too. The rest of the CLI changes are going to be in #606

Issue #608, opened 2022-03-12
Normalizing twarc2 commands


igorbrigadir:
I'm finding that some command line commands are starting to diverge, and become inconsistent. This issue is just for investigating this and coordinating some changes i think we should make. Maybe some breaking changes? It would be great to align all the commands so that they have consistent commonalities and documentation and examples.

To begin with: 

- #599 `name-or-id` pattern is way better than guessing, it should be the same across the others.
- #596
- todo: add other issues here

SamHames:
I'm in favour of this, especially if it means we stop guessing or inferring what was intended.

This is probably enough of a breaking change that we'd want to have at least a major version bump, and ideally some nice managing of errors that we can tell people about the new syntax. This would also let us standardise the handling of usernames, so hopefully we don't have to deal with other instances like #568 and #542.

igorbrigadir:
Maybe all commands that do bulk operations on an input text file vs a single argument should be like `twarc2 bulk followers input.txt output.txt` ? Just another idea.

SamHames:
I like the ID of a bulk subcommand, I can definitely see how that would contribute to consistency: anything that takes an input file is a `bulk` command, everything that takes an argument on the command line is at the top level.

That also gives us a nicer migration path because we can decouple creating new (more consistent) commands under bulk from removing the old commands, allowing for some deprecation warnings or similar.

PR #610, opened 2022-03-17
Add new Quoted tweets endpoint 

igorbrigadir:
There's a new quotes endpoint just launched, this adds the python API and twarc2 command for this. 

https://twittercommunity.com/t/introducing-the-quote-tweets-lookup-endpoint-to-the-twitter-api-v2/168370

### DIFF
--- a/twarc/client2.py
+++ b/twarc/client2.py
@@ -1255,9 +1255,7 @@ class Twarc2:
             if "data" in page:
                 yield page
             else:
-                log.info(
-                    f"Retrieved an empty page of results for quotes of {tweet_id}"
-                )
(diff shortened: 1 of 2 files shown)


SamHames (approved):
:+1:  I saw your initial work and thought I'd missed an announcement :laughing:

igorbrigadir:
No, unfortunately it can only take 1 tweet ID at a time in the command line. But we're gonna change it later in #608 like `twarc2 bulk quotes ids.txt output.jsonl`

Otherwise, you can use the python library to call https://github.com/DocNow/twarc/blob/9f8195415a0387416d3007e302dab83d5015a042/twarc/client2.py#L1231

miguelwon:
Thanks, I will use the library.

PR #628, opened 2022-04-20
Ensure error is raised from _ensure_user when the user doesn't exist

SamHames:
Refactors _ensure_user to not rely on ensure_flattened, and handle
the response directly, as we know exactly what to expect here.


igorbrigadir (approved):

igorbrigadir:
Thanks! Yes, this makes more sense to do it this way!

Issue #629, opened 2022-04-22
twarc1 user_lookup unexpected behavior with string id input


hauselin:
`user_lookup` (twarc1) currently accepts string as input for the `ids` parameter, which can lead to unexpected behavior. 

```py
for u in t.user_lookup("nytimes", id_type="screen_name"):
    print(u["id_str"])
# output
863391
11628
347002675
```

Might be helpful to raise an error if the input is string?

```py
if isinstance(ids, str):
	raise RuntimeError("ids must be an iterable, not string")
```

Or improve the documentation to make clear that `ids` should be an iterable?

SamHames:
Probably a good idea to do both, and the error checking might also make sense in the V2 client as well.

Are you comfortable submitting a PR?

I'd suggest that the error raised should be:
``` python
raise TypeError("ids must be an iterable other than a string")
```

PR #630, opened 2022-04-22
Raise error if user_lookup input is string #629

hauselin:
#629


edsu:
Thanks for reporting this, and for the fix @hauselin!

PR #657, opened 2022-09-30
Add additional expansions and fields for the new tweet edit related API parameters

SamHames:
This PR adds the new field and expansion to capture the full history for edited versions of tweets. Note that the `edit_history_tweet_ids` field is a default field and we don't need to do anything additional to capture this.


igorbrigadir (approved):

igorbrigadir:
Thanks! (Note to self i need to do this for twarc-csv too https://github.com/DocNow/twarc-csv/issues/52 )

PR #660, opened 2022-10-20
bump python version to >=3.6

### LINKED ISSUE #659: “python_requires” should be set with “>=3.6”, but twarc 2.12.0 set as ">=3.3".
Currently, the keyword argument **python_requires** of **setup()** is set as ">=3.3", but I found it is not compatible with **Python<3.6**. My local Python version is 3.5, and I encounter the following error when executing “pip install twarc==2.12.0”

```
Collecting twarc==2.12.0
  Downloading twarc-2.12.0-py3-none-any.whl (60 kB)
     |████████████████████████████████| 60 kB 630 kB/s 
Collecting tqdm>=4.62
  Downloading tqdm-4.64.1-py2.py3-none-any.whl (78 kB)
     |████████████████████████████████| 78 kB 2.1 MB/s 
Collecting click<9,>=7
  Downloading click-7.1.2-py2.py3-none-any.whl (82 kB)
     |████████████████████████████████| 82 kB 1.0 MB/s 
Collecting python-dateutil>=2.8
  Downloading python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
     |████████████████████████████████| 247 kB 4.4 MB/s 
ERROR: Could not find a version that satisfies the requirement humanize>=3.9 (from twarc)
ERROR: No matching distribution found for humanize>=3.9
```

Dependencies of this distribution are listed as follows:

```
click>=7,<9
click-config-file>=0.6
click-plugins>=1
humanize>=3.9
python-dateutil>=2.8
requests_oauthlib>=1.3
tqdm>=4.62
```

I found that **humanize>=3.9** requires **Python>=3.6**, which results in installation failure of twarc in Python3.5.

Way to fix:
modify **setup()** in **setup.py**, amend **python_requires** keyword argument
```
setup(…
     python_requires='>=3.6'
     …)
```
Thanks for your attention.
Best regrads,
PyVCEchecker

igorbrigadir:
fix #659: humanize requires >=3.6 so previous versions would have been broken anyway

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ if __name__ == "__main__":
         classifiers=[
             "License :: OSI Approved :: MIT License",
         ],
-        python_requires=">=3.3",
+        python_requires=">=3.6",
         install_requires=dependencies,
         setup_requires=["pytest-runner"],
         tests_require=[


edsu (approved):
Thanks @igorbrigadir !

PR #664, opened 2022-11-10
implement home timeline reverse chrono

ntorba:
PR for the feature requested here: https://github.com/DocNow/twarc/issues/639

### DIFF
--- a/test_twarc2.py
+++ b/test_twarc2.py
@@ -301,6 +301,20 @@ def atest_timeline():
     assert found >= 200
 
 
+def atest_home_timeline_reverse_chrono():
+    """
+    Test the user timeline endpoints.
+
+    """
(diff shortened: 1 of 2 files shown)


igorbrigadir:
Yeah i'm not sure what happened to the tests either. The reverse chronological timeline does require user context so using `@jack` in the test will always fail: https://developer.twitter.com/en/docs/twitter-api/tweets/timelines/api-reference/get-users-id-reverse-chronological 

Either way, I think it's worth having this because it's hard to get locked (private) accounts otherwise.

edsu:
That was my mistake with `atest_*` that's a dumb hack I sometimes use to just test one thing when I can't be bothered to look up how to do it :-) My apologies!

igorbrigadir:
I'm gonna fix this one up in a bit too!

PR #677, opened 2022-12-20
Add twarc demo video

Quiet27:
I made a short video demonstration of twarc which I thought might be useful to have on the resources page.

In the same PR I updated the screenshot of the Python for Windows installer showing the 'Add Python 3.8 to PATH' option highlighted and ticked. The previous screenshot didn't have it ticked. Even though in text the instructions say to make sure it's ticked, I thought this might be a bit clearer.

I also fixed a link to the resources page from the tutorial page. There might be a better way to do this than putting the full URL to the page, but I wasn't sure.

Thanks for reading!


igorbrigadir:
Awesome! Thanks so much!

igorbrigadir (approved):

edsu:
This is very well done @Quiet27 thank you! I noticed that the video (as downloaded with youtube-dl) is only 20MB. I wonder if we could add it to the docs directory, and then embed it directly in the documentation home page?

Quiet27:
Thanks @edsu! 

More than happy for the YouTube video to be embedded, the following code should do the trick (from the YouTube video > Share > Embed):

```
<iframe width="560" height="315" src="https://www.youtube.com/embed/4DXEeM2AA9Y" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
```

edsu:
Ok, I've added it prominently to the resource documentation. https://twarc-project.readthedocs.io/en/latest/resources/ I thought it might be nice to host the video independently of GOOG but decided to just press the easy button :-)

PR #679, opened 2022-12-20
Add missing variants field to media

### LINKED ISSUE #678: Missing `variants` field in `media.fields`
https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/media 

I somehow missed when they added this, but it has useful alternative formats of media we should be gathering.

igorbrigadir:
Fix #678

Don't think we need a test for this one right? It's just adding a field that was missing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ if __name__ == "__main__":
         setup_requires=["pytest-runner"],
         tests_require=[
             "pytest",
+            "black",
             "pytest-black",
             "python-dotenv",
             "pytz",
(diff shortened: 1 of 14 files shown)


igorbrigadir:
I might go through and quickly check in case there's other fields that should be added

igorbrigadir:
Finally fixed the tests... the issue was the black formatter https://pypi.org/project/black/23.1a1/#history has a pre release version, that `setup.py` prefers to install, while locally, pip only installs pre release versions with `--pre`.

The easier solution is to update your local black version to the latest pre-release:

```
pip install --upgrade --pre black
```

edsu:
Yikes, thank you for catching that atest/test change. That looks like something I committed by mistake!

PR #695, opened 2023-03-20
Disable filter tests

edsu:
It appears that the filter stream just hangs forever now, without returning data. It looks like it is finally disabled? If we don't skip these tests the test suite gets jammed up and never completes.

https://twittercommunity.com/t/announcing-the-deprecation-of-v1-1-statuses-filter-endpoint/182960

### DIFF
--- a/test_twarc.py
+++ b/test_twarc.py
@@ -131,6 +131,7 @@ def test_geocode():
     assert found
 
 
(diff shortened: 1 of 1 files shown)


igorbrigadir (approved):

igorbrigadir:
Yep, also makes sense to finally remove these entirely later.

